PR #11, opened 2021-01-14
Major rewrite of draft-ietf-quic-version-negotiation

### LINKED ISSUE #1: In-line or routing promises vs compatibility
In the common scenario where one has not deployed a new version to 100% of all potential endpoints all the time, the probability of a successful upgrade is p^n, where:
p = probability of landing on server with a proposed version being accepted
n = number of "nonsticky" round trips

If the routing of the packets for an attempted session initiation/connection are routed to the same place, it becomes far more predictable.

All of the above is a long way of saying that compatible is not the same as in-line / routed to the same place.

Providing an in-line upgrade capability allows a reasonably understandable site deployment to do an upgrade without surprising weirdness. Without this, even with "compatible" upgrades, there will be some "fun" deployment issues in practice.

(Issue copied from [individual draft repo](https://github.com/ekr/draft-schinazi-quic-version-negotiation/issues/7), by @grmocg on 2019-11-19)

### LINKED ISSUE #2: Do servers need to understand compatibility?
If there is a definition for a mapping from vM to vN, can a client assume that the server will understand that mapping? Or does the client have to accept an incompatible version negotiation from servers?

It would be nice if compatible mappings could be developed separately from QUIC versions, which suggests that the server can decide whether to apply the mapping.

(Issue copied from [individual draft repo](https://github.com/ekr/draft-schinazi-quic-version-negotiation/issues/6), by @martinthomson on 2019-11-12)

### LINKED ISSUE #3: Comments from Mike Bishop
Comment from @MikeBishop [on the QUIC list](https://mailarchive.ietf.org/arch/msg/quic/0iAgoaoDaKyY_hsSYyyODe8o6HI) on 2019-11-06 :

> Some notes:
>
>  *   Section 3:  “Version negotiation takes place after the retry cycle is over.”  This is only possible for compatible versions; if the server does not support the version the client attempted, it obviously cannot send a Retry packet in that version, because Retry packets are version-specific.
>  *   Section 3:  “Servers must retain the ability to process the Initial packet from older versions….”  No, it means that if the server wants to avoid a VN packet, it needs to be able to process the Initial packet from the version the client tried.
>  *   Section 4:  Rather than putting a count-prefixed list of versions, perhaps consider just embedding the VN packet payload?  It’s equivalent, but easier to explain.
>  *   Section 7:  Clearly it is possible to possible to add new frame types to Initial packets in a future version, I think what you mean to say is that this is a change that would break compatibility between versions.
>
> As a general statement on the document, you assume throughout that all versions have packet types, contain packet types called Initial and Retry, and that Initial packets contain frames.  The odds of those things being true across QUIC versions is likely high, but they’re not invariants, so you can’t assume them here.  You need to state general principles, and then use concepts from QUICv1 as examples of the principle.
>
> For example, rather than discussing version compatibility in terms of Initial packets, you might want to say that versions A and B are “compatible” if the first flight of packets sent by the client in version A carries enough information for the server to generate the corresponding flight in version B, the server speaks version B, and is able to parse at least the first flight of version A.
>
> Also not discussed is the case where the server speaks version(s) which are compatible with what the client offered, but doesn’t speak the version the client offered, even enough to generate the response.  With the right definition of “compatible,” this just falls into the “incompatible” path and works for the server, but you need to call out the fact that the client shouldn’t consider it an error if the server uses the incompatible path between versions that the client thinks are compatible.

### LINKED ISSUE #4: Change framing to improve clarity
Comment from @martinthomson [on the QUIC list](https://mailarchive.ietf.org/arch/msg/quic/x-P5DpFQovPTHi_pkaLVsiC8j10) on 2019-11-04:

> Hi David,
>
> I think that this is broadly the right mechanism.  It would be good to validate it more thoroughly.  
>
> The model and framing could use a little work.
>
> I struggled a little with the framing here because the split between incompatible and compatible versions is very important, but also unclear.  The definition of "compatible" is left for Section 7, but it seems like it should be right up front.
>
> The framing I would choose puts the model up front.  The model that you have have two layers that are applied progressively:
>
> 1. incompatible version negotiation, where the server sends a Version Negotiation packet in response to a version it does not understand.
>
> 2. compatible version negotiation, where the server can generate interpret a version X Initial packet and continue with version Y.
>
> Regarding the model, the definition of "compatible" needs work.
>
> I observe that you only need a mapping from "Initial"-equivalent packets in X to the same in Y - and maybe the ability to generate a Retry in version X (more below) - in order for this to work.  That's where there is a clear functional mapping from a version X packets that initiate the connection to something that fills the same purpose in version Y.  
>
> But the definition you have for "compatible" implies that the mapping is bijective, which I don't think is necessary.  As the draft says, version Y might define (and allow) new frame types in its Initial-equivalent packet(s), but as long as those are either optional or can be synthesized from X, that's OK.  It isn't necessary that every Y have a functional representation in X.  We might not want to define a way to negotiate an older version from a newer one (or at least not for every case).
>
> The ability to generate a Retry stretches this definition a little.  Can you explain why you can't generate a version Y Retry?  I have an inkling, but it's not fully formed.  It complicates the model more than I'd like to have it this way around.

### LINKED ISSUE #5: The client is unable to detect downgrade attacks
While I recognize that it is an explicit design choice of this extension to move all the downgrade detection to the server, it seems unfortunate to prevent the client from detecting the downgrade and instead relying on the server to do so. (I could imagine that some server implementations are lazy/broken and might not perform this detection)?

Alternatively, if the server delivered to the client a set of lists of versions that the server (or the server's peers) might have sent in VN packets, then the client could compare the actual VN packet it received with the elements in the set and detect the case where none of them match. Commonly the server would send one such list, but if it were participating in a staggered rollout, it could send both the "old" and "new" lists?

(Issue copied from [individual draft repo](https://github.com/ekr/draft-schinazi-quic-version-negotiation/issues/3), by @RyanTheOptimist on 2019-11-04)

### LINKED ISSUE #6: Should "Compatible Version" mention 0-RTT?
Consider the case of two difference QUIC versions which are roughly identical in terms of the crypto/handshake/packet layer. But imagine that these two versions, say, renumber streams or some other change which makes 0-RTT data look different. I would assume that such versions would not be "compatible" for the purposes of section 7?

Or alternatively, they could be considered compatible, but the server MUST not process 0-RTT packets, and the client must "unwind" the 0-RTT state (un-open streams, etc)?

(Issue copied from [individual draft repo](https://github.com/ekr/draft-schinazi-quic-version-negotiation/issues/2), by @RyanTheOptimist on 2019-11-04)

### LINKED ISSUE #8: Client Offer Version-Specific Transport Parameters
Related to https://github.com/quicwg/base-drafts/issues/3965.

Should the version negotiation transport parameter allow for (optionally) encoding transport parameters specific to each compatible version? This might improve the range of compatible versions in the future.

For instance, if version 2 changed the meaning of one of the existing transport parameters such that it wasn't directly compatible with version 1's TPs, this would allow the versions themselves to still be considered compatible, because the client could just offer the v2 TPs along with the v1.

### LINKED ISSUE #13: Is the server required to upgrade versions?
Let's assume both client and server support compatible versions `X` and `Y`, and the server prefers `Y` over `X`.
The client now initiates a connection using version `X`, for some reason though the conversion from `X` to `Y` fails.

As the draft is written now, the only way for the server to proceed here is to perform Incomptabile Version Negotiation, i.e. send a VN packet. While this rule makes the process deterministic (and might allow the client to verify the result of the version negotiation, see the open question in https://github.com/quicwg/version-negotiation/issues/5#issuecomment-760623428), it also costs a round trip. When the differences between `X` and `Y` are small, a server might prefer to not bother with upgrading the version, and just establish a connection using `X`. Should we allow that?

### LINKED ISSUE #14: Client's version preference is ignored
Assume that both endpoints support compatible versions `A` and `B`, but the client prefers version `A` over version `B`, while the server prefers version `B` over version `A`.

When the server receives a connection attempt using version `A`, it will detect that the client also supports `B`, and convert it to version `B`. The connection therefore ends up with version `B`.

Is that what we want?  One could argue that the version in use should be controlled by the client, not by the server. This could be easily achieved by only allowing the server to switch to version `B` if the client expressed a preference for `B` over `A`.

DavidSchinazi:
This PR rewrites most of the draft to make it more aligned
with the fact that it's now a stand-alone document that
is not directly tied to QUICv1.

Fixes #1.
Fixes #2.
Fixes #3.
Fixes #4.
Fixes #5.
Fixes #6.
Fixes #8.
Fixes #13.
Fixes #14.

### DIFF
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -362,7 +362,7 @@ In order to facilitate the deployment of future versions of QUIC, designers of
 future versions SHOULD attempt to design their new version such that commonly
 deployed versions are compatible with it. For example, a successor to QUIC
 version 1 may wish to design its transport parameters in a way that does not
-proclude compatibility. Additionally, frames in QUIC version 1 do not use a
+preclude compatibility. Additionally, frames in QUIC version 1 do not use a
 self-describing encoding, so unrecognized frame types cannot be parsed or
 ignored (see the Extension Frames section of {{QUIC}}); this means that new
 versions that wish to be very similar to QUIC version 1 and compatible with it
@@ -421,4 +421,3 @@ Transport Error Codes Registry:
 
 The authors would like to thank Martin Thomson, Mike Bishop, Nick Banks, Ryan
 Hamilton, and Roberto Peon for their input and contributions.
-
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -75,14 +75,6 @@ document are to be interpreted as described in BCP 14 {{!RFC2119}} {{!RFC8174}}
 when, and only when, they appear in all capitals, as shown here.
 
 
-# Connections and Version Negotiation
-
-QUIC connections are shared state between a client and a server {{INV}}. The VN
-mechanism defined in this document operates inside of a QUIC connection; i.e.,
-the packets with the original version are part of the same connection as the VN
-packet and the packets with the negotiated version.
-
-
 # Compatible Versions
 
 If A and B are two distinct versions of QUIC, A is said to be "compatible" with
@@ -97,11 +89,15 @@ compatible with version B and for B not to be compatible with A. This could
 happen for example if version B is a strict superset of version A.
 
 Note that version compatibility does not mean that every single possible
-instance of a first flight will succeed in conversion to the other version. For
-example, if version B is equal to A in all aspects except it introduced a new
-frame in its first flight that version A cannot parse or even ignore, then A
-could still be compatible with B as conversions would succeed for connections
-where that frame is not used.
+instance of a first flight will succeed in conversion to the other version. A
+first flight using version A is said to be "compatible" with version B if two
+conditions are met: first that version A is compatible with version B, and
+second that the conversion of this first flight to version B is well-defined.
+For example, if version B is equal to A in all aspects except it introduced a
+new frame in its first flight that version A cannot parse or even ignore, then
+B could still be compatible with A as conversions would succeed for connections
+where that frame is not used. In this example, first flights using version B
+that carry this new frame would not be compatible with version A.
 
 When a new version of QUIC is defined, it is assumed to not be compatible with
 any other version unless otherwise specified. Similarly, no other version is
@@ -140,6 +136,18 @@ implements enough of the version's specification to parse first flight packets
 but not enough to fully establish a connection using that version.
 
 
+## Connections and Version Negotiation
+
+QUIC connections are shared state between a client and a server {{INV}}. The
+compatible version negotiation mechanism defined in this document (see
+{{compat-vn}}) operates inside of a QUIC connection; i.e., the packets with the
+original version are part of the same connection as the packets with the
+negotiated version. On the other hand, the incompatible version negotiation
+mechanism, which leverages QUIC Version Negotiation packets (see
+{{incompat-vn}}) conceptually operates across two QUIC connections: one before
+the Version Negotiation packet, and a distinct connection after.
+
+
 ## Incompatible Version Negotiation {#incompat-vn}
 
 The server starts incompatible version negotiation by sending a VN packet,
@@ -150,11 +158,6 @@ in the list provided by the server. If it doesn't find one, it aborts the
 connection attempt. Otherwise, it selects a mutually supported version and
 sends a new first flight with that version - we refer to this version as the
 "negotiated version".
-
-When the client sends its new first flight, it SHOULD use the same 5-tuple
-(source and destination IP addresses and UDP port numbers) and the same
-Destination Connection ID as its original first flight. This will increase the
-likelihood of both first flights reaching the same server.
 
 The new first flight will allow the endpoints to establish a connection using
 the negotiated version. The handshake of the negotiated version will exchange
@@ -168,35 +171,36 @@ VN process, see {{downgrade}}.
 When the server can parse the client's first flight using the original version,
 it can extract the client's handshake version information (see
 {{hs-vers-info}}). This contains the list of versions that the client thinks
-its original version is compatible with.
+its first flight is compatible with.
 
 If the server supports one of the client's compatible versions, and the server
 also believes that the original version is compatible with this version, then
-the server attempts to convert the client's first flight to that version. If
-that succeeds, the server replies to the client as if it had received the
-converted first flight. The version used by the server in its reply is refered
-to as the "negotiated version". The server MUST NOT reply with a version that
-is not present in the client's compatible versions, unless it is the original
-version.
+the server converts the client's first flight to that version and replies to
+the client as if it had received the converted first flight. The version used
+by the server in its reply is refered to as the "negotiated version". The
+server MUST NOT reply with a version that is not present in the client's
+compatible versions, unless it is the original version.
 
-If any of these operations fail (except parsing and validation failures which
-are discussed in {{downgrade}}), the server will use the original version if it
-supports it, and if it doesn't then the server will perform incompatible
-version negotiation instead, see {{incompat-vn}}.
+If the server does not find a compatible version, it will use the original
+version if it supports it, and if it doesn't then the server will perform
+incompatible version negotiation instead, see {{incompat-vn}}.
 
 For the duration of the compatible version negotiation process, clients MUST
 use the same 5-tuple (source and destination IP addresses and UDP port
 numbers). During that time, clients MUST also use the same Destination
 Connection ID, except if the server explicitly instructs the client to use a
-different Destination Connection ID. This allows load balancers to ensure that
-packets for a given connection are routed to the same server.
+different Destination Connection ID (for example, a QUIC version 1 server can
+accomplish this by sending an INITIAL packet with a Source Connection ID that
+differed from the client's Destination Connection ID). This allows load
+balancers to ensure that packets for a given connection are routed to the same
+server.
 
 
 # Handshake Version Information {#hs-vers-info}
 
 During the handshake, endpoints will exchange handshake version information,
-which is an opaque blob of data that is defined below. In QUIC version 1, the
-handshake version information is transmitted using a new transport parameter,
+which is a blob of data that is defined below. In QUIC version 1, the handshake
+version information is transmitted using a new transport parameter,
 `version_negotiation`. The contents of handshake version information depend on
 whether the client or server is sending it, and are shown below (using the
 notation from the "Notational Conventions" section of {{QUIC}}):
@@ -231,21 +235,23 @@ all-zeroes.
 Received Negotiation Version Count:
 
 : A variable-length integer specifying the number of Received Negotiation
-  Version fields following it. If the client is sending this first flight in
-  response to a Version Negotiation packet, the subsequent versions SHALL
-  include all the versions from that Version Negotiation packet in order, even
-  if they are not supported by the client (even if the versions are reserved).
-  If the client has not received a Version Negotiation packet on this
-  connection, this field SHALL be 0.
+Version fields following it. If the client is sending this first flight in
+response to a Version Negotiation packet, the subsequent versions SHALL include
+all the versions from that Version Negotiation packet in order, even if they
+are not supported by the client (even if the versions are reserved). If the
+client has not received a Version Negotiation packet on this connection, this
+field SHALL be 0.
 
 Compatible Version Count:
 
 : A variable-length integer specifying the number of Compatible Version fields
-  following it. The client lists all versions that the Currently Attempted
-  Version is compatible with in the subsequent Compatible Version fields,
-  ordered by descending preference. Note that the version in the Currently
-  Attempted Version field MUST be included in the Compatible Version list to
-  allow the client to communicate the currently attempted version's preference.
+following it. The client lists all versions that this first flight is
+compatible with in the subsequent Compatible Version fields, ordered by
+descending preference. Note that the version in the Currently Attempted Version
+field MUST be included in the Compatible Version list to allow the client to
+communicate the currently attempted version's preference. Note that this
+preference is only advisory, servers MAY choose to use their own preference
+instead.
 
 ~~~
 Server Handshake Version Information {
@@ -261,23 +267,24 @@ The content of each field is described below:
 Negotiated Version:
 
 : The version that the server chose to use for the connection. This field MUST
-  be equal to the value of the Version field in the long header that carries
-  this data.
+be equal to the value of the Version field in the long header that carries this
+data.
 
 Supported Version Count:
 
 : A variable-length integer specifying the number of Supported Version fields
-  following it. The server encodes all versions it supports in the subsequent
-  list, ordered by descending preference. Note that the version in the
-  Negotiated Version field MUST be included in the Supported Version list to
-  allow the server to communicate the negotiated version's preference.
-
+following it. The server encodes all versions it supports in the subsequent
+list, ordered by descending preference. Note that the version in the Negotiated
+Version field MUST be included in the Supported Version list to allow the
+server to communicate the negotiated version's preference. Note that this
+preference is only advisory, clients MAY choose to use their own preference
+instead.
 
 Clients MAY include versions following the pattern `0x?a?a?a?a` in their
-Compatible Version list, and the server in their Supported Version list.
+`Compatible Version` list, and the server in their `Supported Version` list.
 Those versions are reserved to exercise version negotiation (see the Versions
 section of {{QUIC}}), and MUST be ignored when parsing these fields. On the
-other hand, the Received Negotiation Version list MUST be identical to the
+other hand, the `Received Negotiation Version` list MUST be identical to the
 received Version Negotiation packet, so clients MUST NOT add or remove reserved
 version from that list.
 
@@ -286,7 +293,8 @@ version from that list.
 
 Clients MUST ignore any received Version Negotiation packets that contain the
 version that they initially attempted. Once a client has reacted to a Version
-Negotiation packet, it MUST drop all subsequent Version Negotiation packets.
+Negotiation packet, it MUST drop all subsequent Version Negotiation packets on
+that connection.
 
 Servers MUST validate that the client's `Currently Attempted Version` matches
 the version in the long header that carried the handshake version information.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-version-negotiation.md
```diff
@@ -269,46 +321,104 @@ initially attempted by default, the client SHOULD cache that information and
...
-When a new version of QUIC is defined, it is assumed to not be compatible
-with any other version unless otherwise specified. Implementations MUST NOT
-assume compatibility between version unless explicitly specified.
+# Considerations for Future Versions
+
+In order to facilitate the deployment of future versions of QUIC, designers of
+future versions SHOULD attempt to design their new version such that commonly
+deployed versions are compatible with it. For example, a successor to QUIC
+version 1 may wish to design its transport parameters in a way that does not
+proclude compatibility. Additionally, frames in QUIC version 1 do not use a
```
nibanks, line 365:
```suggestion
preclude compatibility. Additionally, frames in QUIC version 1 do not use a
```

nibanks (approved):
Generally, looks to be a good rewrite. Thanks!

Beyond the one small nit, I have a question about what document has the responsibility to describe/define what versions are compatible and how. This document by definition is an extension to the QUIC spec. The same may continue to be true for QUIC v2, right? If neither spec directly acknowledge this document, it seems they wouldn't specify how compatibility may work. If so, is it the responsibility of this extension to be updated with all known compatibility rules?

Additionally, would/could we use -29 and v1 as test candidates for this process? Should we define the compatibility rules in this doc?

DavidSchinazi:
@nibanks defining compatible is currently left to future documents. For example QUICv2 could define its compatibility with QUICv1, or it could decide to skip that and rely on a future document.

We could theoretically use draft29 vs QUICv1 as an experiment, but I know we wouldn't implement it because that one is really tricky due to the transport parameter codepoint.

nibanks:
> defining compatible is currently left to future documents

Do we explicitly state this? Should we (can we?) require future documents/versions to respect this process? Even if they're incompatible with other versions, should they reference and require usage of this extension? Would it even still be an extension or directly included?

> could theoretically use draft29 vs QUICv1 as an experiment

I do think it would be a good test case. It might be practically be useful as well, as v1 and -29 start being supported in parallel in the coming months. Perhaps a separate issue should be opened to track this though.

DavidSchinazi:
> > defining compatible is currently left to future documents
> 
> Do we explicitly state this?

We explicitly say that versions are incompatible unless a future document says otherwise.

> Should we (can we?) require future documents/versions to respect this process? Even if they're incompatible with other versions, should they reference and require usage of this extension? Would it even still be an extension or directly included?

We shouldn't require anything. This is just an extension, and we may replace it with a different extension some day.

> > could theoretically use draft29 vs QUICv1 as an experiment
> 
> I do think it would be a good test case. It might be practically be useful as well, as v1 and -29 start being supported in parallel in the coming months. Perhaps a separate issue should be opened to track this though.

A separate issue sounds best, could you file it please?

nibanks:
Ok. Sounds good. #12 opened.

draft-ietf-quic-version-negotiation.md
```diff
@@ -67,105 +67,171 @@ Discussion of this work is encouraged to happen on the QUIC IETF mailing list
...
+When a new version of QUIC is defined, it is assumed to not be compatible with
+any other version unless otherwise specified. Similarly, no other version is
+compatible with the new version unless otherwise specified. Implementations
+MUST NOT assume compatibility between versions unless explicitly specified.
+
+Note that both endpoints might disagree on whether two versions are compatible
+or not. For example, two versions could have been defined concurrently and then
+specified as compatible in a third document much later - in that scenario one
+endpoint might be aware of the compatibility document while the other may not.
+
```
ekr, line 115: Trying to define this as a property of *versions* seems problematic. I think we would be better off with a much simpler model, focuses on flights. A version A first flight is compatible with B if it would be acceptable to a conforming B implementation (this is the situation in TLS, for instance). 

Consider the case where we start with version X.
We now define version Y which has the same initial flight as X except that it requires transport parameter A and requires an initial version of X or Y.

Thus, a version X Initial could consist of:
{version = X, compatible_versions = [X], transport_parameters = [...]}

A compatible version Y would consist of:

At some point in the future, when version X is deprecated, you could send:
{version = Y, compatible_versions = [X, Y], transport_parameters = [...., A]}

This doesn't preclude defining concurrent versions, in that one could could define a version Z which is like X but needs transport parameter B, so it has an X compatible first flight
{version = X, compatible_versions = [X, Z], transport_parameters = [...., B]}

It also has an X, Y, Z compatible first flight:
{version = X, compatible_versions = [X, Y, Z], transport_parameters = [...., A, B]}

This obviously works more cleanly if versions are essentially linear, but works OK if each version is required to accept an Initial which has the external version matching some previous version as well.

draft-ietf-quic-version-negotiation.md
```diff
@@ -67,105 +67,171 @@ Discussion of this work is encouraged to happen on the QUIC IETF mailing list
...
+Upon receiving the VN packet, the client will search for a version it supports
+in the list provided by the server. If it doesn't find one, it aborts the
+connection attempt. Otherwise, it selects a mutually supported version and
+sends a new first flight with that version - we refer to this version as the
+"negotiated version".
+
+When the client sends its new first flight, it SHOULD use the same 5-tuple
+(source and destination IP addresses and UDP port numbers) and the same
+Destination Connection ID as its original first flight. This will increase the
+likelihood of both first flights reaching the same server.
```
ekr, line 157: This SHOULD seems like a problem. If this is the same connection, then this should be a MUST. If not, then DCID should change.

draft-ietf-quic-version-negotiation.md
```diff
@@ -67,105 +67,171 @@ Discussion of this work is encouraged to happen on the QUIC IETF mailing list
...
+## Compatible Version Negotiation {#compat-vn}
+
+When the server can parse the client's first flight using the original version,
+it can extract the client's handshake version information (see
+{{hs-vers-info}}). This contains the list of versions that the client thinks
+its original version is compatible with.
+
+If the server supports one of the client's compatible versions, and the server
+also believes that the original version is compatible with this version, then
+the server attempts to convert the client's first flight to that version. If
```
ekr, line 175: As above, I think this whole "convert" thing is problematic. We should confine ourselves to Initials which are valid for both versions.

draft-ietf-quic-version-negotiation.md
```diff
@@ -67,105 +67,171 @@ Discussion of this work is encouraged to happen on the QUIC IETF mailing list
...
+If any of these operations fail (except parsing and validation failures which
+are discussed in {{downgrade}}), the server will use the original version if it
+supports it, and if it doesn't then the server will perform incompatible
+version negotiation instead, see {{incompat-vn}}.
+
+For the duration of the compatible version negotiation process, clients MUST
+use the same 5-tuple (source and destination IP addresses and UDP port
+numbers). During that time, clients MUST also use the same Destination
+Connection ID, except if the server explicitly instructs the client to use a
+different Destination Connection ID. This allows load balancers to ensure that
```
ekr, line 191: Can you elaborate on how this would happen?

draft-ietf-quic-version-negotiation.md
```diff
@@ -67,105 +67,171 @@ Discussion of this work is encouraged to happen on the QUIC IETF mailing list
...
+packets for a given connection are routed to the same server.
+
+
+# Handshake Version Information {#hs-vers-info}
+
+During the handshake, endpoints will exchange handshake version information,
+which is an opaque blob of data that is defined below. In QUIC version 1, the
```
ekr, line 198: How is this opaque?

draft-ietf-quic-version-negotiation.md
```diff
@@ -225,42 +282,49 @@ received Version Negotiation packet, so clients MUST NOT add or remove reserved
...
-MUST close the connection with an error of type VERSION_NEGOTIATION_ERROR. This
+the version in the long header that carried the handshake version information.
+Similarly, clients MUST validate that the server's `Negotiated Version` matches
+the long header version. If an endpoint's validation fails, it MUST close the
+connection. If the connection was using QUIC version 1, it MUST be closed with
+a transport error of type `VERSION_NEGOTIATION_ERROR`.
+
+When a server parses the client's handshake version information, if the
+`Received Negotiation Version Count` is not zero, the server MUST validate that
+it could have sent the Version Negotation packet described by the client in
```
ekr, line 300: could? Perhaps would?

draft-ietf-quic-version-negotiation.md
```diff
@@ -225,42 +282,49 @@ received Version Negotiation packet, so clients MUST NOT add or remove reserved
...
+If an endpoint receives its peer's Handshake Version Information and fails to
+parse it (for example, if it is too short), then the endpoint MUST close the
+connection. If the connection was using QUIC version 1, it MUST be closed with
+a transport error of type `TRANSPORT_PARAMETER_ERROR`.
 
 
 # Supported Versions
 
-The server's Supported Version list allows it to communicate the full list of
+The server's `Supported Version` list allows it to communicate the full list of
```
ekr, line 327: I don't think this is that helpful and would remove it, but this PR doesn't change that, so we can deal with it separately.

draft-ietf-quic-version-negotiation.md
```diff
@@ -269,46 +333,103 @@ initially attempted by default, the client SHOULD cache that information and
...
+# Interaction with 0-RTT
+
+QUIC version 1 allows sending data from the client to the server during the
+handshake, by using 0-RTT packets. If a future document wishes to define
+compatibility between two versions that support 0-RTT, that document MUST
+address the scenario where there are 0-RTT packets in the client's first
+flight. For example, this could be accomplished by defining which
+transformations are applied to 0-RTT packets. Alternatively, that document
+could specify that compatible version negotiation causes 0-RTT data to be
+rejected by the server.
```
ekr, line 368: This is an interesting point that we didn't really consider in detail with TLS 1.3. However, I think it's off base for the same reason as above. We should just require them to be valid.

ekr (changes requested):

draft-ietf-quic-version-negotiation.md
```diff
@@ -67,105 +67,171 @@ Discussion of this work is encouraged to happen on the QUIC IETF mailing list
...
 document are to be interpreted as described in BCP 14 {{!RFC2119}} {{!RFC8174}}
 when, and only when, they appear in all capitals, as shown here.
 
 
+# Connections and Version Negotiation
+
+QUIC connections are shared state between a client and a server {{INV}}. The VN
+mechanism defined in this document operates inside of a QUIC connection; i.e.,
+the packets with the original version are part of the same connection as the VN
+packet and the packets with the negotiated version.
```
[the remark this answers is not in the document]
DavidSchinazi: We can have it either way. I wrote it as "yes" but we could change that. What do you think?

draft-ietf-quic-version-negotiation.md
DavidSchinazi: Do you have thoughts on what you prefer? I dislike MUST because it can cause issues if version A allows CID lengths of [0-8] and version B only allows [9-20]. (I know this probably won't happen, but I don't want to paint us into that corner)

draft-ietf-quic-version-negotiation.md
DavidSchinazi: What does valid mean? In my understanding of the word, an initial from version A will never be valid in version B because version B requires the version on the long header to be B.

draft-ietf-quic-version-negotiation.md
DavidSchinazi: In QUICv1, the server can select its own CID during the handshake by replying with a source CID that is different from the client's destination CID

draft-ietf-quic-version-negotiation.md
DavidSchinazi: In that versions themselves don't need to worry about what's inside?

draft-ietf-quic-version-negotiation.md
DavidSchinazi: Not really, "would" implies that the server doesn't use randomness when generating VN packets - and we do that for GREASE

draft-ietf-quic-version-negotiation.md
DavidSchinazi: Yeah, please file a separate issue

ekr:
On Mon, Feb 1, 2021 at 4:50 PM David Schinazi <notifications@github.com>
wrote:

> *@DavidSchinazi* commented on this pull request.
> ------------------------------
>
> In draft-ietf-quic-version-negotiation.md
> <https://github.com/quicwg/version-negotiation/pull/11#discussion_r568246064>
> :
>
> > +the negotiated version. The handshake of the negotiated version will exchange
> +handshake version information (see {{hs-vers-info}}) required to ensure that VN
> +was genuine, i.e. that no attacker injected packets in order to influence the
> +VN process, see {{downgrade}}.
> +
> +
> +## Compatible Version Negotiation {#compat-vn}
> +
> +When the server can parse the client's first flight using the original version,
> +it can extract the client's handshake version information (see
> +{{hs-vers-info}}). This contains the list of versions that the client thinks
> +its original version is compatible with.
> +
> +If the server supports one of the client's compatible versions, and the server
> +also believes that the original version is compatible with this version, then
> +the server attempts to convert the client's first flight to that version. If
>
> What does valid mean? In my understanding of the word, an initial from
> version A will never be valid in version B because version B requires the
> version on the long header to be B.
>

Not necessarily. We could say that for version B, the long header could
have either A or B. This basically requires reinterpreting
LongHeader.Version as "format" or "minimum version to interpret" or
something, but it's not inherently a problem.

draft-ietf-quic-version-negotiation.md
DavidSchinazi: @ekr said:
> Not necessarily. We could say that for version B, the long header could have either A or B. This basically requires reinterpreting LongHeader.Version as "format" or "minimum version to interpret" or something, but it's not inherently a problem.

I think we should decouple documents from compatibility. It should be possible to have versions A and B not know about each other, and then someone writes the compatibility doc between them later.

But there are other reasons for things to not be valid: take `h3-32` and `h3-33` for example - they use a different TLS extension codepoints (0xffa5 vs 57) for transport parameters. That makes a first flight from one invalid in the other. However, it's possible to convert from one to the other. That's why I think of conversion instead of validity.

DavidSchinazi:
Merged based on offline conversation with @ekr

PR #42, opened 2021-05-07
Incorporate MT's design

DavidSchinazi:
This PR simplifies the document's design. The new simplified design was by @martinthomson . It removes the indication of server versions that everyone seemed to think was useless, but apart from that has very similar properties to the previous design. One important difference is that it is now the client's responsibility to detect VN-packet-based downgrade attacks instead of the server's.

[An HTML version of what the draft would look like with this PR merged is available here](https://quicwg.org/version-negotiation/simplify/draft-ietf-quic-version-negotiation.html).

### DIFF
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -330,7 +330,7 @@ Both endpoints MUST parse their peer's Version Information during the
 handshake. If the Version Information was missing or if parsing it failed (for
 example, if it is too short or if its length is not divisible by four), then
 the endpoint MUST close the connection; if the connection was using QUIC
-version 1, it MUST be closed with a transport error of type
+version 1, that connection closure MUST use a transport error of type
 `TRANSPORT_PARAMETER_ERROR`.
 
 If a client has reacted to a Version Negotiation packet, it MUST validate that
@@ -339,7 +339,7 @@ version, and that the client would have selected the same negotiated version if
 it had received a Version Negotiation packet whose Supported Versions field had
 the same contents as the server's `Other Versions` field. If any of these
 checks fail, the client MUST close the connection; if the connection was using
-QUIC version 1, it MUST be closed with a transport error of type
+QUIC version 1, that connection closure MUST use a transport error of type
 `VERSION_NEGOTIATION_ERROR`. This prevents an attacker from being able to use
 forged Version Negotiation packets to force a version downgrade.
 
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -74,67 +74,65 @@ While this document mainly discusses a single QUIC server, it is common for
 deployments of QUIC servers to include a fleet of multiple server instances. We
 therefore define the following terms:
 
-Supported Versions:
+Accepted Versions:
 
-: This is the set of versions supported by a given individual server instance.
+: This is the set of versions supported by a given server instance. More
+specifically, these are the versions that a given server instance will use if a
+client sends a first flight using them.
+
+Negotiated Versions:
+
+: This is the set of versions that a given server instance will send in a
+Version Negotiation packet if it receives a first flight from an unknown
+version. This set will most often be equal to the Accepted Versions set, except
+during short transitions while versions are added or removed (see below).
 
 Fully-Deployed Versions:
 
-: This is the set of QUIC versions that is supported by every single QUIC
-server instance in this deployment.
+: This is the set of QUIC versions that is supported and negotiated by every
+single QUIC server instance in this deployment. If a deployment only contains a
+single server instance, then this set is equal to the Negotiated Versions set,
+except during short transitions while versions are added or removed (see below).
 
-Partially-Deployed Versions:
-
-: This is the set of versions supported by a given individual server instance,
-but which aren't supported by every single QUIC server instance in this
-deployment.
-
-If a deployment only contains a single server instance, then the Supported
-Versions is equal to the Fully-Deployed Versions, and the Partially-Deployed
-Versions is the empty set. Note however that during software upgrades these
-fields are upgraded at different times, as described below.
-
-Conversely, if a deployment contains multiple server instances, it is possible
-for software updates to not happen at the exact same time on all server
-instances. In this scenario, some instances will have a different Supported
-Versions from other instances in the same fleet. When that happens, a client
-might receive a Version Negotiation packet from a server instance that doesn't
-support some Partially-Supported Versions; in that case, those
-Partially-Supported Versions won't be listed in the Version Negotiation packet.
+If a deployment contains multiple server instances, it is possible for software
+updates to not happen at the exact same time on all server instances. Because
+of this, a client might receive a Version Negotiation packet from a server
+instance that has already been updated and the client's resulting connection
+attempt might reach a different server instance which hasn't been updated yet.
 
 However, even when there is only a single server instance, it is still possible
 to receive a stale Version Negotiation packet if the server performs its
 software update while the Version Negotiation packet is in flight.
 
-In either of these events, when the client reacts to a Version Negotiation
-packet, it might create a new QUIC connection which could reach a server
-instance which has a different set of Supported Versions than what was received
-in the Version Negotiation packet.
-
 This could cause the version downgrade prevention mechanism described in
 {{downgrade}} to falsely detect a downgrade attack. To avoid that, server
-operators MUST perform a two-step process when they wish to add or remove
+operators SHOULD perform a three-step process when they wish to add or remove
 support for a version:
 
 * When adding support for a new version, the first step is to progressively add
   support for the new version to all server instances. This step updates the
-  Supported Versions but not the Fully-Deployed Versions. Once all server
-  instances have been upgraded, operators wait for at least one minute to allow
-  any in-flight Version Negotiation packets to arrive. Then, the second step is
-  to progressively add the new version to Fully-Deployed Versions on all server
-  instances.
+  Accepted Versions but not the Negotiated Versions nor the Fully-Deployed
+  Versions. Once all server instances have been updated, operators wait for at
+  least one minute to allow any in-flight Version Negotiation packets to
+  arrive. Then, the second step is to progressively add the new version to
+  Negotiated Versions on all server instances. Once complete, operators wait
+  for at least another minute. Finally, the third step is to progressively add
+  the new version to Fully-Deployed Versions on all server instances.
 
 * When removing support for a version, the first step is to progressively
   remove the version from Fully-Deployed Versions on all server instances. Once
   it has been removed on all server instances, operators wait for at least one
   minute to allow any in-flight Version Negotiation packets to arrive. Then,
-  the second step is to progressively remove support for the version from all
-  server instances. That step updates the Supported Versions.
+  the second step is to progressively remove the version from Negotiated
+  Versions on all server instances. Once complete, operators wait for at least
+  another minute. Finally, the third step is to progressively remove support for
+  the version from all server instances. That step updates the Supported
+  Versions.
 
 
-Note that this opens connections to version downgrades during the upgrade
-window, since those could be due to clients communicating with both upgraded
-and non-upgraded server instances.
+Note that this opens connections to version downgrades (but only for
+partially-deployed versions) during the update window, since those could be due
+to clients communicating with both updated and non-updated server instances.
 
 
 # Compatible Versions
@@ -213,7 +211,7 @@ the Version Negotiation packet, and a distinct connection after.
 ## Incompatible Version Negotiation {#incompat-vn}
 
 The server starts incompatible version negotiation by sending a Version
-Negotiation packet. This packet SHALL list the server's set of Fully-Deployed
+Negotiation packet. This packet SHALL list the server's set of Negotiated
 Versions (see {{server-fleet}}) in the packet's Supported Version field. The
 server MAY add reserved versions (as defined in the Versions section of
 {{QUIC}}) to the Supported Version field.
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -340,8 +340,9 @@ it had received a Version Negotiation packet whose Supported Versions field had
 the same contents as the server's `Other Versions` field. If any of these
 checks fail, the client MUST close the connection; if the connection was using
 QUIC version 1, that connection closure MUST use a transport error of type
-`VERSION_NEGOTIATION_ERROR`. This prevents an attacker from being able to use
-forged Version Negotiation packets to force a version downgrade.
+`VERSION_NEGOTIATION_ERROR`. This connection closure prevents an attacker from
+being able to use forged Version Negotiation packets to force a version
+downgrade.
 
 After the process of version negotiation in this document completes, the
 version in use for the connection is the version that the server sent in the
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -67,6 +67,10 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
 document are to be interpreted as described in BCP 14 {{!RFC2119}} {{!RFC8174}}
 when, and only when, they appear in all capitals, as shown here.
 
+In this document, the Maximum Segment Lifetime (MSL) represents the time a QUIC
+packet can exist in the network. Implementations can make this configurable,
+and a RECOMMENDED value is one minute.
+
 
 # Server Deployments of QUIC {#server-fleet}
 
@@ -74,24 +78,24 @@ While this document mainly discusses a single QUIC server, it is common for
 deployments of QUIC servers to include a fleet of multiple server instances. We
 therefore define the following terms:
 
-Accepted Versions:
+Acceptable Versions:
 
 : This is the set of versions supported by a given server instance. More
 specifically, these are the versions that a given server instance will use if a
 client sends a first flight using them.
 
-Negotiated Versions:
+Offered Versions:
 
 : This is the set of versions that a given server instance will send in a
 Version Negotiation packet if it receives a first flight from an unknown
-version. This set will most often be equal to the Accepted Versions set, except
-during short transitions while versions are added or removed (see below).
+version. This set will most often be equal to the Acceptaple Versions set,
+except during short transitions while versions are added or removed (see below).
 
 Fully-Deployed Versions:
 
 : This is the set of QUIC versions that is supported and negotiated by every
 single QUIC server instance in this deployment. If a deployment only contains a
-single server instance, then this set is equal to the Negotiated Versions set,
+single server instance, then this set is equal to the Offered Versions set,
 except during short transitions while versions are added or removed (see below).
 
 If a deployment contains multiple server instances, software updates may not
@@ -112,14 +116,14 @@ support for a version:
 When adding support for a new version:
 
 * The first step is to progressively add support for the new version to all
-  server instances. This step updates the Accepted Versions but not the
-  Negotiated Versions nor the Fully-Deployed Versions. Once all server
-  instances have been updated, operators wait for at least one minute to allow
-  any in-flight Version Negotiation packets to arrive.
+  server instances. This step updates the Acceptable Versions but not the
+  Offered Versions nor the Fully-Deployed Versions. Once all server instances
+  have been updated, operators wait for at least one MSL to allow any in-flight
+  Version Negotiation packets to arrive.
 
-* Then, the second step is to progressively add the new version to Negotiated
+* Then, the second step is to progressively add the new version to Offered
   Versions on all server instances. Once complete, operators wait for at least
-  another minute.
+  another MSL.
 
 * Finally, the third step is to progressively add the new version to
   Fully-Deployed Versions on all server instances.
@@ -128,15 +132,15 @@ When removing support for a version:
 
 * The first step is to progressively remove the version from Fully-Deployed
   Versions on all server instances. Once it has been removed on all server
-  instances, operators wait for at least one minute to allow any in-flight
+  instances, operators wait for at least one MSL to allow any in-flight
   Version Negotiation packets to arrive.
 
-* Then, the second step is to progressively remove the version from Negotiated
+* Then, the second step is to progressively remove the version from Offered
   Versions on all server instances. Once complete, operators wait for at least
-  another minute.
+  another MSL.
 
 * Finally, the third step is to progressively remove support for the version
-  from all server instances. That step updates the Supported Versions.
+  from all server instances. That step updates the Acceptable Versions.
 
 
 Note that this opens connections to version downgrades (but only for
@@ -221,7 +225,7 @@ the Version Negotiation packet, and a distinct connection after.
 
 The server starts incompatible version negotiation by sending a Version
 Negotiation packet. This packet SHALL include each entry from the server's set
-of Negotiated Versions (see {{server-fleet}}) in a Supported Version field. The
+of Offered Versions (see {{server-fleet}}) in a Supported Version field. The
 server MAY add reserved versions (as defined in the Versions section of
 {{QUIC}}) in Supported Version fields.
 
@@ -325,7 +329,7 @@ that connection.
 Both endpoints MUST parse their peer's Version Information during the
 handshake. If the Version Information was missing or if parsing it failed (for
 example, if it is too short or if its length is not divisible by four), then
-the endpoint MUST close the connection. If the connection was using QUIC
+the endpoint MUST close the connection; if the connection was using QUIC
 version 1, it MUST be closed with a transport error of type
 `TRANSPORT_PARAMETER_ERROR`.
 
@@ -334,8 +338,8 @@ the server's `Other Versions` field does not contain the client's original
 version, and that the client would have selected the same negotiated version if
 it had received a Version Negotiation packet whose Supported Versions field had
 the same contents as the server's `Other Versions` field. If any of these
-validations fail, the client MUST close the connection. If the connection was
-using QUIC version 1, it MUST be closed with a transport error of type
+checks fail, the client MUST close the connection; if the connection was using
+QUIC version 1, it MUST be closed with a transport error of type
 `VERSION_NEGOTIATION_ERROR`. This prevents an attacker from being able to use
 forged Version Negotiation packets to force a version downgrade.
 
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -282,9 +282,9 @@ The content of each field is described below:
 
 Chosen Version:
 
-: The version that the sender has chosen to use for this connection. This field
-MUST be equal to the value of the Version field in the long header that carries
-this data.
+: The version that the sender has chosen to use for this connection. In most
+cases, this field will be equal to the value of the Version field in the long
+header that carries this data.
 
 The contents of the `Other Versions` field depends on whether it is sent by the
 client or by the server.
@@ -316,26 +316,30 @@ version that they initially attempted. Once a client has reacted to a Version
 Negotiation packet, it MUST drop all subsequent Version Negotiation packets on
 that connection.
 
-Servers MUST validate that the client's `Chosen Version` matches the version in
-the long header that carried the handshake version information. Similarly,
-clients MUST validate that the server's `Chosen Version` matches the long
-header version. If an endpoint's validation fails, it MUST close the
-connection. If the connection was using QUIC version 1, it MUST be closed with
-a transport error of type `VERSION_NEGOTIATION_ERROR`.
-
-If a client has reacted to a Version Negotiation packet, it MUST parse the
-server's `Other Versions` field and validate that it does not contain the
-client's original version. If this validation fails, the client MUST close the
-connection. If the connection was using QUIC version 1, it MUST be closed with
-a transport error of type `VERSION_NEGOTIATION_ERROR`. This prevents an
-attacker from being able to use forged Version Negotiation packets to force a
-version downgrade.
-
-If an endpoint receives its peer's Handshake Version Information and fails to
-parse it (for example, if it is too short or if its length is not divisible by
+Both endpoints MUST parse their peer's Handshake Version Information during the
+handshake. If the Handshake Version Information was missing or if parsing it
+failed (for example, if it is too short or if its length is not divisible by
 four), then the endpoint MUST close the connection. If the connection was using
 QUIC version 1, it MUST be closed with a transport error of type
 `TRANSPORT_PARAMETER_ERROR`.
+
+If a client has reacted to a Version Negotiation packet, it MUST validate that
+the server's `Other Versions` field does not contain the client's original
+version, and that the client would have selected the same negotiated version if
+it had received a Version Negotiation packet whose Supported Versions field had
+the same contents as the server's `Other Versions` field. If any of these
+validations fail, the client MUST close the connection. If the connection was
+using QUIC version 1, it MUST be closed with a transport error of type
+`VERSION_NEGOTIATION_ERROR`. This prevents an attacker from being able to use
+forged Version Negotiation packets to force a version downgrade.
+
+After the process of version negotiation in this document completes, the
+version in use for the connection is the version that the server sent in the
+`Chosen Version` field of its Handshake Version Information. That remains true
+even if other versions were used in the Version field of long headers at any
+point in the lifetime of the connection; endpoints MUST NOT change the version
+that they consider to be in use based on the Version field of long headers as
+that field could be forged by attackers.
 
 
 # Client Choice of Original Version
@@ -396,9 +400,6 @@ the weakest common version.
 
 The requirement that versions not be assumed compatible mitigates the
 possibility of cross-protocol attacks, but more analysis is still needed here.
-
-The presence of the `Attempted Version` and `Negotiated Version` fields
-mitigates an attacker's ability to forge packets by altering the version.
 
 
 # IANA Considerations
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -68,6 +68,27 @@ document are to be interpreted as described in BCP 14 {{!RFC2119}} {{!RFC8174}}
 when, and only when, they appear in all capitals, as shown here.
 
 
+## Server Deployments of QUIC {#server-terms}
+
+While this document mainly discusses a single QUIC server, it is common for
+deployments of QUIC servers to include a fleet of multiple servers. We
+therefore define the following terms:
+
+Supported Versions:
+
+: This is the set of versions supported by a given individual server.
+
+Fully-Deployed Versions:
+
+: This is the set of QUIC versions that is supported by every single QUIC
+server in this deployment.
+
+Partially-Deployed Versions:
+
+: This is the set of versions supported by a given individual server, but which
+aren't supported by every single QUIC server in this deployment.
+
+
 # Compatible Versions
 
 If A and B are two distinct versions of QUIC, A is said to be "compatible" with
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-version-negotiation.md
```diff
@@ -68,6 +68,27 @@ document are to be interpreted as described in BCP 14 {{!RFC2119}} {{!RFC8174}}
...
 
+## Server Deployments of QUIC {#server-terms}
+
+While this document mainly discusses a single QUIC server, it is common for
+deployments of QUIC servers to include a fleet of multiple servers. We
+therefore define the following terms:
+
+Supported Versions:
+
+: This is the set of versions supported by a given individual server.
```
martinthomson, line 79:
```suggestion
: This is the set of versions supported by a given individual server instance.
```

draft-ietf-quic-version-negotiation.md
```diff
@@ -194,92 +215,49 @@ server.
...
-whether the client or server is sending it, and are shown below (using the
-notation from the "Notational Conventions" section of {{QUIC}}):
+`version_negotiation`. The contents of handshake version information are shown
+below (using the notation from the "Notational Conventions" section of
+{{QUIC}}):
 
 ~~~
-Client Handshake Version Information {
+Handshake Version Information {
   Currently Attempted Version (32),
```
martinthomson, line 224: I think that this might be better named "Chosen Version".  It's not necessarily the first choice, but the choice that is made after seeing what the other side supports (the client chooses by guessing if it doesn't get VN, and chooses from VN if it does; the server chooses from compatible versions).

draft-ietf-quic-version-negotiation.md
```diff
@@ -194,92 +215,49 @@ server.
...
+{: #fig-hvi-format title="Handshake Version Information Format"}
 
 The content of each field is described below:
 
 Currently Attempted Version:
 
-: The version that the client is attempting to use. This field MUST be equal to
-the value of the Version field in the long header that carries this data.
+: This field SHALL be equal to the value of the `Version` field in the long
+header that carries this data.
```
martinthomson, line 235: I'm not sure that this is the definition we want.  I *think* that you are right that this matches the version in the long header, but that is an accident of the design.  The key is that this is the version that the endpoint chose.

It might be possible with retransmissions of Initial packets to include this information even after the server - and server - has switched to a compatible version.  While it might be obvious that the server got the info, the retransmission machinery on the client might not have received an ACK and so it could send this again.

draft-ietf-quic-version-negotiation.md
```diff
@@ -194,92 +215,49 @@ server.
...
-all-zeroes.
+Client-Sent Other Versions:
 
-Received Negotiation Version Count:
+: When sent by a client, the `Other Versions` field lists all the versions that
+this first flight is compatible with, ordered by descending preference. Note
+that the version in the `Currently Attempted Version` field MUST be included in
+this list to allow the client to communicate the currently attempted version's
+preference. Note that this preference is only advisory, servers MAY choose to
+use their own preference instead.
```
martinthomson, line 247: I'm still not on board with preference ordering.  But let's track that separately.

draft-ietf-quic-version-negotiation.md
```diff
@@ -194,92 +215,49 @@ server.
...
-client has not received a Version Negotiation packet on this connection, this
-field SHALL be 0.
+Server-Sent Other Versions:
 
-Compatible Version Count:
+: When sent by a server, the `Other Versions` field lists all the
+Fully-Supported Versions of this server deployment, see {{server-terms}}. This
+field conveys Fully-Supported Versions instead of Supported Versions to allow
+scenarios where the client's first flight (before receiving VN) reached a
+different server than the client's subsequent flight (after receiving VN).
```
martinthomson, line 255: You might need to expand on this, as it is a key concept.  

This might be worth an entire section that says something like:

The client might receive a Version Negotiation packet from a server instance that doesn't support some Partially-Supported Versions; if it does so, those Partially-Supported Versions won't be listed in the Version Negotiation packet.  As part of validating the server's Other Versions, the client verifies that it would not have chosen any of the versions the server listed.  If the client supports any Partially-Supported Version and would prefer that version, it will detect a downgrade attack.  To avoid this, the server does not advertise any version that is Partially-Supported.

(And then we need text about how to turn on new versions properly.  (1) Deploy code and accept connections; (2) turn on VN for that version; (3) turn on Other Versions for that version and enjoy downgrade protection.)

draft-ietf-quic-version-negotiation.md
```diff
@@ -194,92 +215,49 @@ server.
...
-`Compatible Version` list, and the server in their `Supported Version` list.
-Those versions are reserved to exercise version negotiation (see the Versions
-section of {{QUIC}}), and will be ignored when parsing these fields. On the
-other hand, the `Received Negotiation Version` list MUST be identical to the
-received Version Negotiation packet, so clients MUST NOT add or remove reserved
-version from that list.
+Clients and servers MAY both include versions following the pattern
+`0x?a?a?a?a` in their `Other Versions` list. Those versions are reserved to
+exercise version negotiation (see the Versions section of {{QUIC}}), and will
+be ignored when parsing these fields.
```
martinthomson, line 260: These aren't ignored, it's just that they are never selected.

draft-ietf-quic-version-negotiation.md
```diff
@@ -296,42 +274,19 @@ the long header version. If an endpoint's validation fails, it MUST close the
...
-to start enforcing that the new version is present in `Received Negotiation
-Versions`. This opens connections to version downgrades during the upgrade
-window, since those could be due to clients communicating with both upgraded
-and non-upgraded servers.
-
-If an endpoint receives its peer's Handshake Version Information and fails to
-parse it (for example, if it is too short), then the endpoint MUST close the
+If a client has reacted to a Version Negotiation packet, it MUST parse the
+server's `Other Versions` field and validate that it does not contain the
+client's original version. If this validation fails, the server MUST close the
```
martinthomson, line 279:
```suggestion
client's original version. If this validation fails, the client MUST close the
```

draft-ietf-quic-version-negotiation.md
```diff
@@ -296,42 +274,19 @@ the long header version. If an endpoint's validation fails, it MUST close the
...
 connection. If the connection was using QUIC version 1, it MUST be closed with
-a transport error of type `TRANSPORT_PARAMETER_ERROR`.
+a transport error of type `VERSION_NEGOTIATION_ERROR`. This mitigates an
+attacker's ability to forge Version Negotiation packets to force a version
+downgrade.
```
martinthomson, line 283:
```suggestion
a transport error of type `VERSION_NEGOTIATION_ERROR`. This prevents 
an attacker from being able to use forged Version Negotiation packets to
force a version downgrade.
```

martinthomson (commented):
Much easier than I had expected, which is good.

There's more work to do here, and I have only reviewed the diff, but this is good.

draft-ietf-quic-version-negotiation.md
DavidSchinazi: Agreed, done.

draft-ietf-quic-version-negotiation.md
DavidSchinazi: I filed #43 to track this.

draft-ietf-quic-version-negotiation.md
DavidSchinazi: Agreed, fleshed out the "Server Deployments of QUIC" section

draft-ietf-quic-version-negotiation.md
DavidSchinazi: Fixed

DavidSchinazi (commented):
Thanks for the review!

draft-ietf-quic-version-negotiation.md
DavidSchinazi: Agreed, I tweaked the text.

draft-ietf-quic-version-negotiation.md
```diff
@@ -210,6 +189,54 @@ balancers to ensure that packets for a given connection are routed to the same
...
+but which aren't supported by every single QUIC server instance in this
+deployment.
+
+If a deployment only contains a single server instance, then the Supported
+Versions is equal to the Fully-Deployed Versions, and the Partially-Deployed
+Versions is the empty set.
+
+Conversely, if a deployment contains multiple server instances, it is possible
+for software updates to not happen at the exact same time on all server
+instances. In this scenario, some instances will have a different Supported
```
martinthomson, line 219: I think that you can just say that an instantaneous upgrade is impossible.  A server deployment that includes multiple server instances cannot upgrade all instances instantaneously.  Even if that were possible, there could still be outdated Version Negotiation packets in flight that might result in negotiation failure if downgrade protection were engaged atomically.

This also applies to a single instance deployment.  Though you might avoid the problem by ensuring that your upgrade involves enough downtime to avoid the in-flight VN problem, you might as well just apply the same logic as a multi-server deployment.

Based on that I would remove the "single server instance" paragraph and replace it with a description of the in-flight VN problem.

draft-ietf-quic-version-negotiation.md
```diff
@@ -210,6 +189,54 @@ balancers to ensure that packets for a given connection are routed to the same
...
+might receive a Version Negotiation packet from a server instance that doesn't
+support some Partially-Supported Versions; in that case, those
+Partially-Supported Versions won't be listed in the Version Negotiation packet.
+When the client reacts to that Version Negotiation packet, it might create a
+new QUIC connection which could be routed to a different server instance, which
+has a different set of Supported Versions. The version downgrade prevention
+mechanism described in {{downgrade}} is robust to this scenario and will not
+falsely detect a downgrade attack.
+
+If a server operator is progressively deploying a new QUIC version throughout
```
martinthomson, line 230: Based on the above, I would remove the condition here: just describe how a new version is deployed.

draft-ietf-quic-version-negotiation.md
```diff
@@ -221,18 +248,19 @@ below (using the notation from the "Notational Conventions" section of
...
 The content of each field is described below:
 
-Currently Attempted Version:
+Chosen Version:
 
-: This field SHALL be equal to the value of the `Version` field in the long
-header that carries this data.
+: The version that the sender has chosen to use for this connection. This field
+MUST be equal to the value of the Version field in the long header that carries
+this data.
```
martinthomson, line 263: As I said before, I think that you can - at best - say that the field is generally equal to the Version field in the packet that carries the data.  You can't use "MUST".

draft-ietf-quic-version-negotiation.md
```diff
@@ -241,23 +269,20 @@ Client-Sent Other Versions:
 
 : When sent by a client, the `Other Versions` field lists all the versions that
```
martinthomson, line 270: I just noticed this editorial convention: I think that you should avoid using backticks for field names.  The QUIC spec doesn't use them.  (One nasty consequence of using them is that you get scare quotes in the text rendering.)

draft-ietf-quic-version-negotiation.md
```diff
@@ -210,6 +189,54 @@ balancers to ensure that packets for a given connection are routed to the same
...
+has a different set of Supported Versions. The version downgrade prevention
+mechanism described in {{downgrade}} is robust to this scenario and will not
+falsely detect a downgrade attack.
+
+If a server operator is progressively deploying a new QUIC version throughout
+its fleet, it MUST perform a two-step process where it first progressively adds
+support for the new version. This adds the new version to Supported Versions
+but not to Fully-Deployed Versions. Once all server instances have been
+upgraded, the second step is to progressively add the version to Fully-Deployed
+Versions on all server instances. This opens connections to version downgrades
```
martinthomson, line 235: This needs to be three steps:

1. Add code for version X
2. Advertise version X in Version Negotiation
3. Add version X to Other Versions

The reason I say this is the in-flight VN.  Let's say that you support version X and you just add it both VN and OV.  What about clients that 2ms ago got a VN without X in it?

The transition between step 2 and 3 can be short.  You can run that on a fairly short timer for a single server deployment, but you need to ensure that all VNs are out of the system before you move to step 3 or connections will fail.

The logic here is:

* don't update the advisory signals (VN) until the code is in place
* don't update the downgrade protection (OV) until all the advisory signals are in place

Rollback is similar: first remove downgrade protection, then remove advisory signals, then remove actual support.

One thing I like about this design is that we because we don't rely on VN being correct, the design can be made robust against VN being wrong for a short period.

See also https://quicwg.org/ops-drafts/draft-ietf-quic-applicability.html#name-enabling-new-versions

martinthomson (commented):
I realize now that the 2-step upgrade might need to be 3-step.  See below for more.

draft-ietf-quic-version-negotiation.md
martinthomson: So the server doesn't need to validate what the client provides (it is picking from the list, and **use** >> **validation**).

The client verifies that (a) the version it sent prior to receiving a Version Negotiation packet is not listed, and (b) it would not have selected a different version if it knew what versions the server truly supports.  Note here that the client's choice is based on what the server signals in other versions, the chosen version (if we drop ordering, as we should).

I realize now that the server can't drop compatible versions from this list.  There's a long and complicated argument for why that doesn't work if you are interested, but I need to eat a well-overdue lunch and it might be better over a video link because it's tricky.

DavidSchinazi: > So the server doesn't need to validate what the client provides (it is picking from the list, and **use** >> **validation**).

I've tweaked the text to clarify that this is the version used, not the one from long headers. That allowed me to drop this validation.

> The client verifies that (a) the version it sent prior to receiving a Version Negotiation packet is not listed, and (b) it would not have selected a different version if it knew what versions the server truly supports.

Oh good catch I was missing (b).

> Note here that the client's choice is based on what the server signals in other versions, the chosen version (if we drop ordering, as we should).

Note that in the current proposal we have ordering of the client's versions but not of the server's versions.

> I realize now that the server can't drop compatible versions from this list. There's a long and complicated argument for why that doesn't work if you are interested, but I need to eat a well-overdue lunch and it might be better over a video link because it's tricky.

In the current proposal endpoints can disagree on which versions are compatible so we need them in the list in case the client doesn't think they are compatible.

martinthomson: That is not the reason I had, but as the server generates its list after the client, it would know what the client thinks is compatible when generating its list.

DavidSchinazi: Oh that's a good point. I'd love to hear the full story over VC one of these days then, but no rush.

But overall, do you think this PR is ready to be merged? (If you do I'll email the list to get WG input)

draft-ietf-quic-version-negotiation.md
```diff
@@ -68,6 +68,75 @@ document are to be interpreted as described in BCP 14 {{!RFC2119}} {{!RFC8174}}
...
+operators MUST perform a two-step process when they wish to add or remove
+support for a version:
+
+* When adding support for a new version, the first step is to progressively add
+  support for the new version to all server instances. This step updates the
+  Supported Versions but not the Fully-Deployed Versions. Once all server
+  instances have been upgraded, operators wait for at least one minute to allow
+  any in-flight Version Negotiation packets to arrive. Then, the second step is
+  to progressively add the new version to Fully-Deployed Versions on all server
+  instances.
```
martinthomson, line 125: Do you add the code AND start sending VN with the new version at the same time?  Because sending VN in that state will result in connection failures.  As I said, this probably needs a three step dance.

draft-ietf-quic-version-negotiation.md
```diff
@@ -194,92 +266,47 @@ server.
 During the handshake, endpoints will exchange handshake version information,
 which is a blob of data that is defined below. In QUIC version 1, the handshake
 version information is transmitted using a new transport parameter,
-`version_negotiation`. The contents of handshake version information depend on
-whether the client or server is sending it, and are shown below (using the
-notation from the "Notational Conventions" section of {{QUIC}}):
+`version_negotiation`. The contents of handshake version information are shown
+below (using the notation from the "Notational Conventions" section of
+{{QUIC}}):
```
martinthomson, line 271: You can use `{{Section 15 of QUIC}}` if you like, or even `{{Section 15 (Versions) of QUIC}}`.  Both work (if you have a recent kramdown version).

draft-ietf-quic-version-negotiation.md
```diff
@@ -194,92 +266,47 @@ server.
...
 
 ~~~
-Client Handshake Version Information {
-  Currently Attempted Version (32),
-  Previously Attempted Version (32),
-  Received Negotiation Version Count (i),
-  Received Negotiation Version (32) ...,
-  Compatible Version Count (i),
-  Compatible Version (32) ...,
+Handshake Version Information {
```
martinthomson, line 274: How about just "Version Information"?

martinthomson (approved):
I think that your upgrade process is busted, but that can be fixed later if you want to open a separate issue.

draft-ietf-quic-version-negotiation.md
DavidSchinazi: No, the key property here is that VN packets carry the Fully-Deployed Versions, so that happens after the 1min wait. Or am I missing something?

draft-ietf-quic-version-negotiation.md
DavidSchinazi: `{{Section 1.3 of QUIC}}` works but `{{Section 1.3 (Notational Conventions) of QUIC}}` doesn't. I'd rather have the section title when the other document isn't yet published because it could get renumbered.

FWIW it's failing with `Error: IDREF attribute target references an unknown ID "Section_1.3__Notational_Conventions__of_QUIC", at None` and I'm at https://github.com/martinthomson/i-d-template/commit/4639e5ce1002e49ee93a4bb3d9acd7644b3a83de.

draft-ietf-quic-version-negotiation.md
DavidSchinazi: I like that, done.

draft-ietf-quic-version-negotiation.md
DavidSchinazi: Oh I'm realizing you said a recent kramdown, not a recent i-d-template. I'll look into updating that.

draft-ietf-quic-version-negotiation.md
```diff
@@ -68,6 +68,75 @@ document are to be interpreted as described in BCP 14 {{!RFC2119}} {{!RFC8174}}
...
+server instance in this deployment.
+
+Partially-Deployed Versions:
+
+: This is the set of versions supported by a given individual server instance,
+but which aren't supported by every single QUIC server instance in this
+deployment.
+
+If a deployment only contains a single server instance, then the Supported
+Versions is equal to the Fully-Deployed Versions, and the Partially-Deployed
```
RyanTheOptimist, line 93: nit: "Supported Versions" is plural, but "is" is singular. You might consider using "are" instead of "is" or alternatively dodging by using "Supported Versions set" (which is singular).

draft-ietf-quic-version-negotiation.md
```diff
@@ -68,6 +68,75 @@ document are to be interpreted as described in BCP 14 {{!RFC2119}} {{!RFC8174}}
...
+Versions is the empty set. Note however that during software upgrades these
+fields are upgraded at different times, as described below.
```
RyanTheOptimist, line 95: nit: "fields" or "sets"

draft-ietf-quic-version-negotiation.md
martinthomson: So the problem is that if you change VN and Other Versions at the same time. You will have a mismatch between VN and Other Versions in both directions.  Some clients will see VN without the new version and OV with.  Other clients will see VN with the new version and OV without.

It's that former of those two that creates a problem; if the client sees VN without version X and then sees version X in OV, then it will fail the connection.

DavidSchinazi: /facepalm. Yup, you're right. Three steps it is.

draft-ietf-quic-version-negotiation.md
```diff
@@ -289,49 +313,30 @@ version that they initially attempted. Once a client has reacted to a Version
...
+If a client has reacted to a Version Negotiation packet, it MUST validate that
+the server's `Other Versions` field does not contain the client's original
+version, and that the client would have selected the same negotiated version if
+it had received a Version Negotiation packet whose Supported Versions field had
+the same contents as the server's `Other Versions` field. If any of these
+validations fail, the client MUST close the connection. If the connection was
+using QUIC version 1, it MUST be closed with a transport error of type
+`VERSION_NEGOTIATION_ERROR`. This prevents an attacker from being able to use
+forged Version Negotiation packets to force a version downgrade.
+
```
huitema, line 332: There is a tradeoff here between "punish the attacker" and "punish the client". Suppose that when all is said and done, the server proposes version X. Suppose also that version X is equally or better preferred by the client over the original version Y and the version Z proposed after version negotiation. For all practical matter, the attack has been thwarted, and the negotiation landed in a good place. Do you really want the client to discard the connection? Isn't that providing VN packets with the power of DOS?

huitema: On the other hand, I do like putting the onus of verification on the client. That's a significant improvement over the previous draft version.

draft-ietf-quic-version-negotiation.md
```diff
@@ -289,49 +313,30 @@ version that they initially attempted. Once a client has reacted to a Version
...
+After the process of version negotiation in this document completes, the
+version in use for the connection is the version that the server sent in the
+`Chosen Version` field of its Version Information. That remains true even if
+other versions were used in the Version field of long headers at any point in
+the lifetime of the connection; endpoints MUST NOT change the version that they
+consider to be in use based on the Version field of long headers as that field
+could be forged by attackers.
```
huitema, line 340: Yes! That does make the code simpler.

huitema (commented):
Almost there, but I think we need clarity on the clients verification-and-drop requirements.

draft-ietf-quic-version-negotiation.md
martinthomson: I would be ok with saying that the client rejects the connection according to its policy. If it is content, then that is fine. Though I would strongly encourage a default posture of rejection. There are many other ways that an on path attack at this stage can be used to deny service, but leaving the protocol open to downgrade means that it is harder to defend. What if the perfectly acceptable version is broken, unbeknownst to the client?

draft-ietf-quic-version-negotiation.md
DavidSchinazi: Thanks :)

draft-ietf-quic-version-negotiation.md
DavidSchinazi: I agree with MT here, VN can already cause DoS (just send an empty VN packet) so I'd rather be more conservative unless we have a specific scenario we want to enable.

draft-ietf-quic-version-negotiation.md
```diff
@@ -68,6 +68,73 @@ document are to be interpreted as described in BCP 14 {{!RFC2119}} {{!RFC8174}}
...
+
+Fully-Deployed Versions:
+
+: This is the set of QUIC versions that is supported and negotiated by every
+single QUIC server instance in this deployment. If a deployment only contains a
+single server instance, then this set is equal to the Negotiated Versions set,
+except during short transitions while versions are added or removed (see below).
+
+If a deployment contains multiple server instances, it is possible for software
+updates to not happen at the exact same time on all server instances. Because
```
MikeBishop, line 98:
```suggestion
updates to happen a few server instances at a time. Because
```
Removes a split infinitive and language that feels excessively informal ("exact same time").

draft-ietf-quic-version-negotiation.md
```diff
@@ -68,6 +68,73 @@ document are to be interpreted as described in BCP 14 {{!RFC2119}} {{!RFC8174}}
...
+
+* When adding support for a new version, the first step is to progressively add
+  support for the new version to all server instances. This step updates the
+  Accepted Versions but not the Negotiated Versions nor the Fully-Deployed
+  Versions. Once all server instances have been updated, operators wait for at
+  least one minute to allow any in-flight Version Negotiation packets to
+  arrive. Then, the second step is to progressively add the new version to
+  Negotiated Versions on all server instances. Once complete, operators wait
+  for at least another minute. Finally, the third step is to progressively add
+  the new version to Fully-Deployed Versions on all server instances.
```
MikeBishop, line 120:
```suggestion
support for a version.

When adding support for a new version:

- The first step is to progressively add
  support for the new version to all server instances. This step updates the
  Accepted Versions but not the Negotiated Versions nor the Fully-Deployed
  Versions. Once all server instances have been updated, operators wait for at
  least one minute to allow any in-flight Version Negotiation packets to
  arrive. 
- Next, progressively add the new version to
  Negotiated Versions on all server instances. Once complete, operators wait
  for at least another minute.
- Finally, progressively add
  the new version to Fully-Deployed Versions on all server instances.
```
Modulo rewrapping, I think moving the bullets down a level is clearer.

draft-ietf-quic-version-negotiation.md
```diff
@@ -68,6 +68,73 @@ document are to be interpreted as described in BCP 14 {{!RFC2119}} {{!RFC8174}}
...
+
+* When removing support for a version, the first step is to progressively
+  remove the version from Fully-Deployed Versions on all server instances. Once
+  it has been removed on all server instances, operators wait for at least one
+  minute to allow any in-flight Version Negotiation packets to arrive. Then,
+  the second step is to progressively remove the version from Negotiated
+  Versions on all server instances. Once complete, operators wait for at least
+  another minute. Finally, the third step is to progressively remove support for
+  the version from all server instances. That step updates the Supported
+  Versions.
```
MikeBishop, line 130:
```suggestion
When removing support for a version:

- The first step is to progressively
  remove the version from Fully-Deployed Versions on all server instances. Once
  it has been removed on all server instances, operators wait for at least one
  minute to allow any in-flight Version Negotiation packets to arrive.
 - Next, progressively remove the version from Negotiated
  Versions on all server instances. Once complete, operators wait for at least
  another minute.
- Finally, progressively remove support for
  the version from all server instances. That step updates the Supported
  Versions.
```
...and the same down here.

draft-ietf-quic-version-negotiation.md
```diff
@@ -143,8 +210,11 @@ the Version Negotiation packet, and a distinct connection after.
 
 ## Incompatible Version Negotiation {#incompat-vn}
 
-The server starts incompatible version negotiation by sending a VN packet,
-listing all the versions that it does support.
+The server starts incompatible version negotiation by sending a Version
+Negotiation packet. This packet SHALL list the server's set of Negotiated
+Versions (see {{server-fleet}}) in the packet's Supported Version field. The
```
MikeBishop, line 215: Nit:  Supported Version doesn't contain a list -- it contains a single version and repeats as many times as necessary to convey the list.

```suggestion
Negotiation packet. This packet SHALL include each entry from the server's
set of Negotiated Versions (see {{server-fleet}}) in a Supported Version
field. The
```

draft-ietf-quic-version-negotiation.md
```diff
@@ -154,17 +224,17 @@ sends a new first flight with that version - we refer to this version as the
 
 The new first flight will allow the endpoints to establish a connection using
 the negotiated version. The handshake of the negotiated version will exchange
-handshake version information (see {{hs-vers-info}}) required to ensure that VN
-was genuine, i.e. that no attacker injected packets in order to influence the
-VN process, see {{downgrade}}.
+Version Information (see {{vers-info}}) required to ensure that VN was genuine,
```
MikeBishop, line 227: Capitalizing this suggests it's referring to the structure, but it's the information within that's relevant.
```suggestion
information (see {{vers-info}}) used to ensure that VN was genuine,
```

draft-ietf-quic-version-negotiation.md
```diff
@@ -154,17 +224,17 @@ sends a new first flight with that version - we refer to this version as the
...
+see {{downgrade}}.
 
 
 ## Compatible Version Negotiation {#compat-vn}
 
 When the server can parse the client's first flight using the original version,
-it can extract the client's handshake version information (see
-{{hs-vers-info}}). This contains the list of versions that the client thinks
-its first flight is compatible with.
+it can extract the client's Version Information (see {{vers-info}}). This
```
MikeBishop, line 235:
```suggestion
it can extract the client's Version Information structure (see {{vers-info}}). This
```

draft-ietf-quic-version-negotiation.md
```diff
@@ -189,97 +259,51 @@ balancers to ensure that packets for a given connection are routed to the same
...
-which is a blob of data that is defined below. In QUIC version 1, the handshake
-version information is transmitted using a new transport parameter,
-`version_negotiation`. The contents of handshake version information depend on
-whether the client or server is sending it, and are shown below (using the
-notation from the "Notational Conventions" section of {{QUIC}}):
+During the handshake, endpoints will exchange Version Information, which is a
+blob of data that is defined below. In QUIC version 1, the Version Information
+is transmitted using a new transport parameter, `version_information`. The
+contents of Version Information are shown below (using the notation from the
+"Notational Conventions" section of {{QUIC}}):
```
MikeBishop, line 268:
```suggestion
contents of Version Information are shown below (using the notation from
{{Section 1.3 of QUIC}}):
```

draft-ietf-quic-version-negotiation.md
```diff
@@ -189,97 +259,51 @@ balancers to ensure that packets for a given connection are routed to the same
...
-Clients MAY include versions following the pattern `0x?a?a?a?a` in their
-`Compatible Version` list, and the server in their `Supported Version` list.
-Those versions are reserved to exercise version negotiation (see the Versions
-section of {{QUIC}}), and will be ignored when parsing these fields. On the
-other hand, the `Received Negotiation Version` list MUST be identical to the
-received Version Negotiation packet, so clients MUST NOT add or remove reserved
-version from that list.
+Clients and servers MAY both include versions following the pattern
+`0x?a?a?a?a` in their `Other Versions` list. Those versions are reserved to
+exercise version negotiation (see the Versions section of {{QUIC}}), and will
```
MikeBishop, line 305:
```suggestion
exercise version negotiation (see {{Section 15 of QUIC}}), and will
```

MikeBishop (approved):

draft-ietf-quic-version-negotiation.md
DavidSchinazi: Fixed

draft-ietf-quic-version-negotiation.md
DavidSchinazi: Fixed

draft-ietf-quic-version-negotiation.md
DavidSchinazi: Fixed.

draft-ietf-quic-version-negotiation.md
DavidSchinazi: Fixed

draft-ietf-quic-version-negotiation.md
DavidSchinazi: Fixed

draft-ietf-quic-version-negotiation.md
```diff
@@ -68,6 +68,82 @@ document are to be interpreted as described in BCP 14 {{!RFC2119}} {{!RFC8174}}
 when, and only when, they appear in all capitals, as shown here.
 
 
+# Server Deployments of QUIC {#server-fleet}
+
+While this document mainly discusses a single QUIC server, it is common for
+deployments of QUIC servers to include a fleet of multiple server instances. We
+therefore define the following terms:
+
+Accepted Versions:
```
ekr, line 77: Nit: I would call this "acceptable"

draft-ietf-quic-version-negotiation.md
```diff
@@ -68,6 +68,82 @@ document are to be interpreted as described in BCP 14 {{!RFC2119}} {{!RFC8174}}
...
+
+: This is the set of versions supported by a given server instance. More
+specifically, these are the versions that a given server instance will use if a
+client sends a first flight using them.
+
+Negotiated Versions:
```
ekr, line 83: And I would call this "offered versions"

draft-ietf-quic-version-negotiation.md
```diff
@@ -68,6 +68,82 @@ document are to be interpreted as described in BCP 14 {{!RFC2119}} {{!RFC8174}}
...
+
+* The first step is to progressively add support for the new version to all
+  server instances. This step updates the Accepted Versions but not the
+  Negotiated Versions nor the Fully-Deployed Versions. Once all server
+  instances have been updated, operators wait for at least one minute to allow
+  any in-flight Version Negotiation packets to arrive.
+
+* Then, the second step is to progressively add the new version to Negotiated
+  Versions on all server instances. Once complete, operators wait for at least
+  another minute.
```
ekr, line 122: This should be phrased in terms of MSL.

draft-ietf-quic-version-negotiation.md
```diff
@@ -68,6 +68,82 @@ document are to be interpreted as described in BCP 14 {{!RFC2119}} {{!RFC8174}}
...
+
+* Finally, the third step is to progressively add the new version to
+  Fully-Deployed Versions on all server instances.
+
+When removing support for a version:
+
+* The first step is to progressively remove the version from Fully-Deployed
+  Versions on all server instances. Once it has been removed on all server
+  instances, operators wait for at least one minute to allow any in-flight
+  Version Negotiation packets to arrive.
```
ekr, line 132: What does this mean operationally. Fully-Deployed is defined as being across all the servers.

draft-ietf-quic-version-negotiation.md
```diff
@@ -289,49 +322,30 @@ version that they initially attempted. Once a client has reacted to a Version
...
-notified of more recent versions the server supports. If the client notices
-that the server supports a version that is more preferable that the one
-initially attempted by default, the client SHOULD cache that information and
-attempt the preferred version in subsequent connections.
+Both endpoints MUST parse their peer's Version Information during the
+handshake. If the Version Information was missing or if parsing it failed (for
+example, if it is too short or if its length is not divisible by four), then
+the endpoint MUST close the connection. If the connection was using QUIC
+version 1, it MUST be closed with a transport error of type
+`TRANSPORT_PARAMETER_ERROR`.
```
ekr, line 330: Doesn't this mean you can't do incompatible VN when you offer VN=1?

draft-ietf-quic-version-negotiation.md
```diff
@@ -289,49 +322,30 @@ version that they initially attempted. Once a client has reacted to a Version
...
+
+If a client has reacted to a Version Negotiation packet, it MUST validate that
+the server's `Other Versions` field does not contain the client's original
+version, and that the client would have selected the same negotiated version if
+it had received a Version Negotiation packet whose Supported Versions field had
+the same contents as the server's `Other Versions` field. If any of these
+validations fail, the client MUST close the connection. If the connection was
```
ekr, line 337:
```suggestion
checks fail, the client MUST close the connection. If the connection was
```

draft-ietf-quic-version-negotiation.md
```diff
@@ -289,49 +322,30 @@ version that they initially attempted. Once a client has reacted to a Version
...
+using QUIC version 1, it MUST be closed with a transport error of type
+`VERSION_NEGOTIATION_ERROR`. This prevents an attacker from being able to use
+forged Version Negotiation packets to force a version downgrade.
```
ekr, line 340: Why?

draft-ietf-quic-version-negotiation.md
DavidSchinazi: The moment you remove it from the list it means that server no longer sends it in the TP.

draft-ietf-quic-version-negotiation.md
DavidSchinazi: Clarified this by using a semi-colon.

draft-ietf-quic-version-negotiation.md
DavidSchinazi: Is it unclear? Can you suggest text?

draft-ietf-quic-version-negotiation.md
ekr: How is that different from removal from Negotiated versions

draft-ietf-quic-version-negotiation.md
```diff
@@ -289,49 +326,30 @@ version that they initially attempted. Once a client has reacted to a Version
...
-notified of more recent versions the server supports. If the client notices
-that the server supports a version that is more preferable that the one
-initially attempted by default, the client SHOULD cache that information and
-attempt the preferred version in subsequent connections.
+Both endpoints MUST parse their peer's Version Information during the
+handshake. If the Version Information was missing or if parsing it failed (for
+example, if it is too short or if its length is not divisible by four), then
+the endpoint MUST close the connection; if the connection was using QUIC
+version 1, it MUST be closed with a transport error of type
+`TRANSPORT_PARAMETER_ERROR`.
```
ekr, line 334:
```suggestion
the endpoint MUST close the connection with a version-specific error; if the connection was using QUIC
version 1, that error MUST be `TRANSPORT_PARAMETER_ERROR`.
```

draft-ietf-quic-version-negotiation.md
DavidSchinazi: Negotiated Versions is sent in VN packets, these are sent in the TP. They're different.

draft-ietf-quic-version-negotiation.md
ekr: This too is about the binding of "This". In this case, it's closing the connection that prevents the VN packets forcing a downgrade. I think the text I proposed above would work.

PR #49, opened 2021-07-16
Client validation

martinthomson:
In light of discussion in slack, the problem here is that the validation
is based on the *negotiated* version.  Validation needs to be based on
the client's *selected* version.

I've also moved text about validating lack of support for the original
version to the incompatible VN section, alongside a reminder to validate
connection IDs.  Security does depend on this, but the client ignores
these bad VN packets.  That provides more effective defense against
attack than what was here originally.

### DIFF
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -28,6 +28,9 @@ author:
     organization: Mozilla
     email: ekr@rtfm.com
 
+normative:
+  RFC8999:
+
 
 --- abstract
 
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -229,12 +229,12 @@ that follows.
 The server starts incompatible version negotiation by sending a Version
 Negotiation packet. This packet SHALL include each entry from the server's set
 of Offered Versions (see {{server-fleet}}) in a Supported Version field. The
-server MAY add reserved versions (as defined in the Versions section of
-{{QUIC}}) in Supported Version fields.
+server MAY add reserved versions (as defined in {{Section 6.3 of QUIC}}) in
+Supported Version fields.
 
 A client MUST ignore a Version Negotiation packet if it contains the original
 version attempted by the client.  The client also ignores a Version Negotiation
-that contains incorrect connection ID fields.
+that contains incorrect connection ID fields; see {{Section 6 of RFC8999}}.
 
 Upon receiving the VN packet, the client will search for a version it supports
 in the list provided by the server. If it doesn't find one, it aborts the
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -235,7 +235,7 @@ of Offered Versions (see {{server-fleet}}) in a Supported Version field. The
 server MAY add reserved versions (as defined in {{Section 6.3 of QUIC}}) in
 Supported Version fields.
 
-A client MUST ignore a Version Negotiation packet if it contains the original
+Clients will ignore a Version Negotiation packet if it contains the original
 version attempted by the client.  The client also ignores a Version Negotiation
 that contains incorrect connection ID fields; see {{Section 6 of RFC8999}}.
 
@@ -352,7 +352,7 @@ transport error of type `VERSION_NEGOTIATION_ERROR`.
 The client MUST validate the server `Other Versions` field by confirming that it
 would have attempted the same version with this knowledge of the versions the
 server supports. That is, the client would have selected the same version if it
-received Version Negotiation packet that listed the versions in the server's
+received a Version Negotiation packet that listed the versions in the server's
 `Other Versions` field, plus the negotiated version. If the client would have
 selected a different version, the client MUST close the connection; if the
 connection was using QUIC version 1, that connection closure MUST use a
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-version-negotiation.md
```diff
@@ -28,6 +28,9 @@ author:
     organization: Mozilla
     email: ekr@rtfm.com
 
+normative:
```
DavidSchinazi, line 31: Out of curiosity, does a syntax like `{{Section 6 of !RFC8999}}` not work?

draft-ietf-quic-version-negotiation.md
```diff
@@ -229,8 +232,12 @@ that follows.
 The server starts incompatible version negotiation by sending a Version
 Negotiation packet. This packet SHALL include each entry from the server's set
 of Offered Versions (see {{server-fleet}}) in a Supported Version field. The
-server MAY add reserved versions (as defined in the Versions section of
-{{QUIC}}) in Supported Version fields.
+server MAY add reserved versions (as defined in {{Section 6.3 of QUIC}}) in
+Supported Version fields.
+
+A client MUST ignore a Version Negotiation packet if it contains the original
```
DavidSchinazi, line 238: I'm not a huge fan of repeating normative statements from a normative reference. How about something non-normative that's backed up by the reference:
```suggestion
Clients will ignore a Version Negotiation packet if it contains the original
```

draft-ietf-quic-version-negotiation.md
```diff
@@ -342,16 +349,20 @@ the Version Information was missing, the client MUST close the connection; if
...
-the same contents as the server's `Other Versions` field. If any of these
-checks fail, the client MUST close the connection; if the connection was using
-QUIC version 1, that connection closure MUST use a transport error of type
-`VERSION_NEGOTIATION_ERROR`. This connection closure prevents an attacker from
-being able to use forged Version Negotiation packets to force a version
-downgrade.
+The client MUST validate the server `Other Versions` field by confirming that it
+would have attempted the same version with this knowledge of the versions the
+server supports. That is, the client would have selected the same version if it
+received Version Negotiation packet that listed the versions in the server's
```
DavidSchinazi, line 355:
```suggestion
received a Version Negotiation packet that listed the versions in the server's
```

draft-ietf-quic-version-negotiation.md
```diff
@@ -342,16 +349,20 @@ the Version Information was missing, the client MUST close the connection; if
...
-the server's `Other Versions` field does not contain the client's original
-version, and that the client would have selected the same negotiated version if
-it had received a Version Negotiation packet whose Supported Versions field had
-the same contents as the server's `Other Versions` field. If any of these
-checks fail, the client MUST close the connection; if the connection was using
-QUIC version 1, that connection closure MUST use a transport error of type
-`VERSION_NEGOTIATION_ERROR`. This connection closure prevents an attacker from
-being able to use forged Version Negotiation packets to force a version
-downgrade.
+The client MUST validate the server `Other Versions` field by confirming that it
```
DavidSchinazi, line 352: This requirement only applies to clients that have reacted to a VN packet right? What validation MUST a client do if this connection didn't involve a VN packet? (in other words, why drop the "If a client has reacted to a Version Negotiation packet," part?)

DavidSchinazi (commented):
Thanks for writing this up!

Also, can you please add Anthony Rossi to the Acknowledgements since he pointed out this issue?

draft-ietf-quic-version-negotiation.md
anrossi: Why shouldn't the client not do the same validation during compatible version negotiation?

draft-ietf-quic-version-negotiation.md
martinthomson: It didn't when I last touched that code.

draft-ietf-quic-version-negotiation.md
martinthomson: There are ways in which the two are different, so I think David is right here.

Let's say that there are two groups of versions, the versions in each group compatible with each other, but incompatible with the versions in the other group.  The client prefers a version in the first group, but the client also knows that the versions in the first group are not widely implemented.  As the client doesn't like the added latency of version negotiation, it chooses the second group, which contains a version that is totally OK.  However, if it gets a Version Negotiation packet as a result, it will happily switch to the first group.

So yeah, let's go with the "if VN" condition.

draft-ietf-quic-version-negotiation.md
```diff
@@ -342,17 +349,23 @@ the Version Information was missing, the client MUST close the connection; if
...
 transport error of type `VERSION_NEGOTIATION_ERROR`.
 
-If a client has reacted to a Version Negotiation packet, it MUST validate that
-the server's `Other Versions` field does not contain the client's original
-version, and that the client would have selected the same negotiated version if
-it had received a Version Negotiation packet whose Supported Versions field had
-the same contents as the server's `Other Versions` field. If any of these
-checks fail, the client MUST close the connection; if the connection was using
-QUIC version 1, that connection closure MUST use a transport error of type
+If the client received and acted on a Version Negotiation packet, it client MUST
```
DavidSchinazi, line 352:
```suggestion
If the client received and acted on a Version Negotiation packet, the client MUST
```

draft-ietf-quic-version-negotiation.md
```diff
@@ -342,17 +349,23 @@ the Version Information was missing, the client MUST close the connection; if
...
+validate the server `Other Versions` field.  The `Other Versions` field is
```
DavidSchinazi, line 353:
```suggestion
validate the server's `Other Versions` field.  The `Other Versions` field is
```

DavidSchinazi (approved):
Thanks! Approved modulo two nits

PR #50, opened 2021-07-19
Tweak the intro

### LINKED ISSUE #29: Make introduction match the text
The introduction reads like the draft only defines compatible version negotiation.  It probably needs to be written to be more general.

martinthomson:
The "single round trip" thing isn't really a guarantee given various
factors.  Setting this up as *additional* round trips is cleaner and
more accurate.  I adjusted the paragraph structure to match.

Closes #29.


DavidSchinazi (approved):
Thanks, that is a lot clearer

PR #65, opened 2021-11-22
adding example scenarios

### LINKED ISSUE #55: Backward compatibility with the version negotiation of RFC 9000
Suppose that a client supports QUIC v1, v2 and the new negotiation while a server supports RFC 9000 only. 

The following normal scenario succeeds because the client does not receive a VN packet so the client does not have to check `version_information`:

```
- C: Initial: ver=1, version_information = 1, {2, 1}
- S: Initial: ver=1, Handshake
- C: Handshake: ver=1
```

However, the following scenario, I think, fails:

```
- C: Initial: ver=greasing, version_information = greasing, {2, 1, greasing}
- S: VN: ver=0, {1}
- C: Initial: ver=1, version_information = 1, {1}
- S: Initial: ver=1, Handshake
- C: Handshake: CC
```

The client sent CC because `version_information` is missing.

The latter is the common method to test the version negotiation of RFC 9000. Are we giving up this method?

kazu-yamamoto:
This should close #55.

### DIFF
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -498,6 +498,44 @@ codepoint in the 0-63 range to replace the provisional codepoint described above
 
 --- back
 
+# Example scenarios
+
+The following is an example of compatible negotiation in the case where both endpoints support both QUIC version 2 and the version negotiation. C and S denotes client and server, respectively.
+
+~~~
+C: Initial: ver=1, version_information = 1, {2, 1}
+S: Initial: ver=2; Handshake: ver=2, version_information = 2, {2, 1}
+C: Handshake: ver=2
+~~~
+
+The following is an example of incompatible negotiation in the case where both endpoints support both QUIC version 2 and the version negotiation.
+
+~~~
+C: Initial: ver=greasing, version_information = greasing, {2, 1, greasing}
+S: Version Negotiation: ver=0, {2, 1}
+C: Initial: ver=2, version_information = 2, {2, 1}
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-version-negotiation.md
```diff
@@ -498,6 +498,44 @@ codepoint in the 0-63 range to replace the provisional codepoint described above
...
+~~~
+C: Initial: ver=1, version_information = 1, {2, 1}
+S: Initial: ver=1; Handshake: ver=1
+C: Handshake: ver=1
+~~~
+
+The following is an example of version negotiation failure in the case where a client supports QUIC version 2 and the version negotiation and a server supports RFC 9000 only. Since the client receives a version negotiation packet, version_information is necessary in the Handshake packet from a server for the version negotiation.
+
+~~~
+C: Initial: ver=greasing, version_information = greasing, {2, 1, greasing}
```
DavidSchinazi, line 532: This example uses a greasing version, and we never specified how those are used. Should we remove that example? What purpose does this example serve?

draft-ietf-quic-version-negotiation.md
```diff
@@ -498,6 +498,44 @@ codepoint in the 0-63 range to replace the provisional codepoint described above
 
 --- back
 
+# Example scenarios
+
+The following is an example of compatible negotiation in the case where both endpoints support both QUIC version 2 and the version negotiation. C and S denotes client and server, respectively.
```
DavidSchinazi, line 503: QUIC version 2 doesn't exist yet. Should we use a fake version for these examples?

DavidSchinazi:
Closing now that we've resolved #55 differently

PR #66, opened 2021-11-24
Special case QUIC version 1

### LINKED ISSUE #55: Backward compatibility with the version negotiation of RFC 9000

DavidSchinazi:
Fixes #55

### DIFF
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -482,9 +482,9 @@ from the server's transport parameters, then the client SHALL proceed as if the
 server's transport parameters contained a version_information transport
 parameter with a Chosen Version set to 0x00000001 and an Other Version list
 containing exactly one version set to 0x00000001. This allows version
-negotiation to work with servers that only support QUIC version 1. Note
-that implementations which wish to use VN to negotiate versions other
-than QUIC version 1 will need to implement this draft.
+negotiation to work with servers that only support QUIC version 1. Note that
+implementations which wish to use version negotiation to negotiate versions
+other than QUIC version 1 will need to implement this draft.
 
 
 # Security Considerations
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -417,6 +417,21 @@ applied to 0-RTT packets. Alternatively, that document could specify that
 compatible version negotiation causes 0-RTT data to be rejected by the server.
 
 
+# Special Handling for QUIC Version 1
+
+Because QUIC version 1 was the only IETF Standards Track version of QUIC
+published before this document, it is handled specially as follows: if a client
+is starting a QUIC version 1 connection in response to a received Version
+Negotiation packet, and the version_information transport parameter is missing
+from the server's transport parameters, then the client SHALL proceed as if the
+server's transport parameters contained a version_information transport
+parameter with a Chosen Version set to 0x00000001 and an Other Version list
+containing exactly one version set to 0x00000001. This allows version
+negotiation to work with servers that only support QUIC version 1, while
+preventing downgrade attacks against servers that support QUIC version 1 and
+earlier versions.
+
+
 # Considerations for Future Versions
 
 In order to facilitate the deployment of future versions of QUIC, designers of
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -485,9 +485,9 @@ from the server's transport parameters, then the client SHALL proceed as if the
 server's transport parameters contained a version_information transport
 parameter with a Chosen Version set to 0x00000001 and an Other Version list
 containing exactly one version set to 0x00000001. This allows version
-negotiation to work with servers that only support QUIC version 1, while
-preventing downgrade attacks against servers that support QUIC version 1 and
-earlier versions.
+negotiation to work with servers that only support QUIC version 1. Note
+that implementations which wish to use VN to negotiate versions other
+than QUIC version 1 will need to implement this draft.
 
 
 # Security Considerations


ekr:
I'm not sure that this is the right angle. Instead, I would suggest that we simply say that VN can't be used to negotiate pre-v1 versions of QUIC, which seems to achieve the same objective.

martinthomson:
It's true that you could avoid using any sort of version negotiation signaling to select a version that doesn't support this extension (with the exception of version 1).  That still leaves a client that supports a new version and version 1 without protection when it has to validate the response from a server that only supports version 1.  This change addresses the problem by saying that if you don't support this mechanism properly, then you must only support version 1.

We could *also* say that Version Negotiation packets cannot be used to select draft versions of QUIC, but I don't think that makes any real difference.

ekr:
> It's true that you could avoid using any sort of version negotiation signaling to select a version that doesn't support this extension (with the exception of version 1). That still leaves a client that supports a new version and version 1 without protection when it has to validate the response from a server that only supports version 1. This change addresses the problem by saying that if you don't support this mechanism properly, then you must only support version 1.

I don't think I'm following you here. If the client supports version N > 1 and the server only supports 1, then what are you trying to protect against? You can only negotiate 1.

martinthomson:
If the client supports this specification, it is required to fail to connect in that case because the transport parameter is absent.  That's what this change is addressing.

ekr:
Hmm, well that's a matter of interop, not protection. So, I agree some special casing is needed, but I don't think this is necessarily it. I would just say that if the parameter is not present, then you can still negotiate v1, rather than trying to pretend it sent a parameter it didn't.

martinthomson:
Same net effect.  Sounds like that comes down to editorial preference.

martinthomson:
...oh, I just realized that this isn't completely true.  The formulation with `1, {1}` says something more: it says that you don't support any other version.  That might be important for those cases where you might have other incompatible versions available.

ekr:
```Authors:
    D. Schinazi
    Google LLC
    E. Rescorla
    Mozilla
```

I guess David and I will discuss.

ekr:
> ...oh, I just realized that this isn't completely true. The formulation with `1, {1}` says something more: it says that you don't support any other version. That might be important for those cases where you might have other incompatible versions available.

Yes, and I'm proposing we don't address that case, because future versions should use this mechanism and we just want to get rid of past versions.

martinthomson:
But would it be reasonable to require that - if you implement a new version - you implement this extension for version 1?

ekr:
Yes, I think we should require that.

martinthomson:
This change is consistent with that decision: if a server implements an incompatible version X, then it will be required to send `1, {1, X}`, which will allow the client to confirm that it would not have chosen X over 1.  That provides the client with protection against a downgrade to version 1 from X.  If you merely allow the client to negotiate version X, then that client gets no downgrade protection in that case.

ekr:
As above, I propose to merely forbid negotiating any older version at all
via this mechanism.

DavidSchinazi:
@ekr to clarify your position: in the scenario where
- the client supports version 1, version 2, and this extension
- the client supports only version 1 and not this extension
- the client starts with v2
- server replies with a VN packet that only contains v1
- client restarts with v1 but doesn't get version_information from the server

Are you suggesting that in this scenario the client should fail the handshake?

ekr:
On Wed, Nov 24, 2021 at 11:48 AM David Schinazi ***@***.***>
wrote:

> @ekr <https://github.com/ekr> to clarify your position: in the scenario
> where
>
>    - the client supports version 1, version 2, and this extension
>    - the client supports only version 1 and not this extension
>
> I assume this sentence should say "server"?


>    - the client starts with v2
>    - server replies with a VN packet that only contains v1
>    - client restarts with v1 but doesn't get version_information from the
>    server
>
> Are you suggesting that in this scenario the client should fail the
> handshake?
>

No. I believe this should succeed. What I believe should fail is that the
client supports V2, V1, and some old (preV1) draft version, and server only
supports that draft version and they use this extension to negotiate that.

DavidSchinazi:
> What I believe should fail is that the client supports V2, V1, and some old (preV1) draft version, and server only supports that draft version and they use this extension to negotiate that.

@ekr in this scenario did the server send the version_information transport parameter?

ekr:
Doesn't matter. I think we should actively forbid negotiation of pre-V1 versions, regardless of whether they implement this mechanism or not.

DavidSchinazi:
I think that's outside the purview of this draft. Google's going to keep using Google QUIC and h3-29 until our aggregated data shows it's safe to remove support for those versions - we're not going to treat those versions different from QUICv1 until then.

ekr:
It's not even remotely outside of the purview of this draft. For example, RFC 8996 forbids negotiating TLS 1.0. 

As for what Google chooses to do or not to do, as is often said, there are no protocol police. But that doesn't really have that much to do with whether the IETF should endorse removing these versions.

Regardless, I don't really see what the issue is here: presumably the problem is old *clients*, who will not be implementing this draft in any case, so you can simply accept them as a server, but not do VN for them.

DavidSchinazi:
I'm failing to understand how your proposal is an alternative to this PR. Consider the scenario I mentioned in https://github.com/quicwg/version-negotiation/pull/66#issuecomment-978173869 - in the current version of the draft, that scenario fails whereas with this PR that scenario succeeds. Since that scenario only involves QUICv1 and QUICv2, how are pre-QUICv1 versions relevant?

ekr:
I'll write something up in more detail.

kazu-yamamoto:
@ekr Gentle ping.

kazu-yamamoto:
I have implemented this PR feature in my client and confirmed that the backward compatibility is maintained.

kazu-yamamoto:
@ekr Would you write up?

martinduke:
I don't see why this is a special case for QUICv1. It should apply to all single-version servers.

I thought our position was that any multiple-version endpoint needs to support the VN draft. If a server was e.g. v2 only for some silly reason, it could omit the TP with no risk of downgrade. The client should interpret lack of the TP as a single-version server, whether it's v1 or not.

The alternative is to make the TP mandatory for all non-v1 servers. I could edit the v2 draft accordingly if that's the consensus.

martinduke:
I don't see why this is a special case for QUICv1. It should apply to all single-version servers.

I thought our position was that any multiple-version endpoint needs to support the VN draft. If a server was e.g. v2 only for some silly reason, it could omit the TP with no risk of downgrade. The client should interpret lack of the TP as a single-version server, whether it's v1 or not.

The alternative is to make the TP mandatory for all non-v1 servers. I could edit the v2 draft accordingly if that's the consensus.

DavidSchinazi:
"it could omit the TP with no risk of downgrade" is not correct. You can have an attack where the client thinks it's speaking one version and the server is thinking it's speaking another version because an attacker is rewriting packets as they flow by. This can't happen with v1 vs v2 because we changed the TLS HKDF labels but it's a risk in general. Keeping the Chosen Version in the transport parameters avoids this attack and simplifies security analysis

martinduke:
> "it could omit the TP with no risk of downgrade" is not correct. You can have an attack where the client thinks it's speaking one version and the server is thinking it's speaking another version because an attacker is rewriting packets as they flow by. This can't happen with v1 vs v2 because we changed the TLS HKDF labels but it's a risk in general. Keeping the Chosen Version in the transport parameters avoids this attack and simplifies security analysis

Doesn't that collapse once the Version field is properly authenticated?

And anyway 9001 says we SHOULD roll new labels for each version, FWIW

DavidSchinazi:
> Doesn't that collapse once the Version field is properly authenticated?

The Version field is not authenticated if it's not in the TLS transcript.

> And anyway 9001 says we SHOULD roll new labels for each version, FWIW

That doesn't apply to non-TLS versions

draft-ietf-quic-version-negotiation.md
```diff
@@ -475,6 +475,21 @@ applied to 0-RTT packets. Alternatively, that document could specify that
...
+published before this document, it is handled specially as follows: if a client
+is starting a QUIC version 1 connection in response to a received Version
+Negotiation packet, and the version_information transport parameter is missing
+from the server's transport parameters, then the client SHALL proceed as if the
+server's transport parameters contained a version_information transport
+parameter with a Chosen Version set to 0x00000001 and an Other Version list
+containing exactly one version set to 0x00000001. This allows version
+negotiation to work with servers that only support QUIC version 1, while
+preventing downgrade attacks against servers that support QUIC version 1 and
+earlier versions.
```
ekr, line 490:
```suggestion
containing exactly one version set to 0x00000001. This allows version
negotiation to work with servers that only support QUIC version 1, while
preventing downgrade attacks against servers that support QUIC version 1 and
earlier versions.
```
```suggestion
containing exactly one version set to 0x00000001. This allows version
negotiation to work with servers that only support QUIC version 1. Note
that implementations which wish to use VN to negotiate versions other
than QUIC version 1 will need to implement this draft.
```

I believe this is more correct.

My preferred text would actually be phrased differently, which is to say:
"from the server's transport parameters, then the client SHALL omit
the checks specified in Section XXX". I believe this has the same impact
but is less prescriptive about the details. However, @martinthomson's comments
convinced me that this is just a matter of spelling.

ekr, line 490:
```suggestion
containing exactly one version set to 0x00000001. This allows version
negotiation to work with servers that only support QUIC version 1, while
preventing downgrade attacks against servers that support QUIC version 1 and
earlier versions.
```
```suggestion
containing exactly one version set to 0x00000001. This allows version
negotiation to work with servers that only support QUIC version 1. Note
that implementations which wish to use VN to negotiate versions other
than QUIC version 1 will need to implement this draft.
```

I believe this is more correct.

My preferred text would actually be phrased differently, which is to say:
"from the server's transport parameters, then the client SHALL omit
the checks specified in Section XXX". I believe this has the same impact
but is less prescriptive about the details. However, @martinthomson's comments
convinced me that this is just a matter of spelling.

DavidSchinazi: Thanks. Committed that suggestion for now, happy to tweak editorial details later

PR #82, opened 2022-03-04
Clarify minimum levels of compliance

### LINKED ISSUE #80: Differing levels of support for the standard
As discussed in https://github.com/quicwg/quic-v2/issues/25, there are various levels of partial support for this standard, different for client and server, and there ought to be normative language around these capabilities. There are certain modes that MUST be supported along with multiple versions, and others that are just performance improvements.

AFAICT, for any server that supports >1 versions, there are two possible levels of support:
1. Send the TP and support incompatible negotiation
2. Support compatible negotiation

For a client, it's more complicated: @DavidSchinazi proposes:
1. No support: do not the send nor parse the TP. If you get a VN packet, fail the QUIC connection
2. Downgrade prevention only: send the TP with only one other version, also parse and validate the TP. If you get a VN packet, fail the QUIC connection. This protects against intermediaries that are rewriting the version field on Initial packets. In v1/v2, these attacks will be detected by the authenticated version field in subsequent handshake packets, but that is not generalizable to all versions of QUIC.
3. Compatible only: downgrade prevention and be prepared to accept an upgrade
4. Incompatible only: advertise only one version, so there's downgrade prevention when trying again with new version after a VN packet.
5. Full: downgrade prevention, compatible VN and incompatible VN

martinduke:
Fixes #80

### DIFF
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -418,9 +418,10 @@ version that is compatible with a version advertised in the transport parameter,
 it MUST support compatible version negotiation.
 
 Therefore, clients can only avoid support for compatible negotiation by either
-(1) never attempting connection based on a Version Negotiation packet, and
-electing not to include the transport parameter; or (2) never including a
-compatible version in the trnasport parameter.
+
+* never attempting connection based on a Version Negotiation packet, and
+electing not to include the transport parameter; or
+* never including a compatible version in the trnasport parameter.
 
 # Client Choice of Original Version
 
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -387,6 +387,40 @@ lifetime of the connection; endpoints MUST NOT change the version that they
 consider to be in use based on the Version field of long headers as that field
 could be forged by attackers.
 
+# Minimum support
+
+Hosts can support this standard incrementally.
+
+{{RFC8999}} and {{RFC9000}} already describe how servers send Version
+Negotiation packets in response to long header packets with unsupported
+version fields, even if they support only one version. Doing so prevents long
+timeouts at the client, but servers might not send the packet multiple times
+for the same connection, or when under stress.
+
+If a server supports two or more QUIC versions and sends Version Negotiation
+packets, it MUST send the Version Information Transport Parameter {{vers-info}},
+and parse the client's transport parameter in accordance with {{downgrade}}.
+Failure to comply would leave the connection vulnerable to downgrade attacks, or
+cause connection failure, if the client initiates Incompatible Version
+Negotiation.
+
+A server that supports two or more compatible QUIC versions MAY implement
+Compatible version negotiation if it prefers one version over another.
+
+A client that supports two or more versions, and chooses to respond to a Version
+Negotiation packet with a connection attempt using one of the advertised
+versions, MUST include a Version Information Transport Parameter and process the
+server's transport parameter in accordance with {{downgrade}}.
+
+If a client sends a Version Information Transport Parameter in a packet using a
+version that is compatible with a version advertised in the transport parameter,
+it MUST support compatible version negotiation.
+
+Therefore, clients can only avoid support for compatible negotiation by either
+(1) never attempting connection based on a Version Negotiation packet, and
+electing not to include the transport parameter; or (2) never including a
+supported version in that transport parameter with which the one in use is
+compatible.
 
 # Client Choice of Original Version
 
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -389,7 +389,8 @@ could be forged by attackers.
 
 # Minimum support
 
-Hosts can support this standard incrementally.
+Hosts can support this standard incrementally to simplify implementation, if
+not all features of the design are desired.
 
 {{INV}} and {{QUIC}} already describe how servers send Version Negotiation
 packets in response to long header packets with unsupported version fields, even
@@ -398,14 +399,14 @@ but servers might not send the packet multiple times for the same connection, or
 when under stress.
 
 If a server supports two or more QUIC versions and sends Version Negotiation
-packets, it MUST send the Version Information Transport Parameter {{vers-info}},
-and parse the client's transport parameter in accordance with {{downgrade}}.
+packets, it MUST follow the requirements in {{vers-info}} and {downgrade}}.
 Failure to comply would leave the connection vulnerable to downgrade attacks, or
 cause connection failure, if the client initiates Incompatible Version
-Negotiation.
+Negotiation ({{incompat-vn}}).
 
 A server that supports two or more compatible QUIC versions MAY implement
-Compatible version negotiation if it prefers one version over another.
+Compatible version negotiation ({{compat-vn}}) if it prefers one version over
+another.
 
 A client that supports two or more versions, and chooses to respond to a Version
 Negotiation packet with a connection attempt using one of the advertised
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-version-negotiation.md
```diff
@@ -387,6 +387,40 @@ lifetime of the connection; endpoints MUST NOT change the version that they
...
+if they support only one version. Doing so prevents long timeouts at the client,
+but servers might not send the packet multiple times for the same connection, or
+when under stress.
+
+If a server supports two or more QUIC versions and sends Version Negotiation
+packets, it MUST send the Version Information Transport Parameter {{vers-info}},
+and parse the client's transport parameter in accordance with {{downgrade}}.
+Failure to comply would leave the connection vulnerable to downgrade attacks, or
+cause connection failure, if the client initiates Incompatible Version
+Negotiation.
```
DavidSchinazi, line 405: nit: add reference to that section

draft-ietf-quic-version-negotiation.md
```diff
@@ -387,6 +387,40 @@ lifetime of the connection; endpoints MUST NOT change the version that they
...
+
+A server that supports two or more compatible QUIC versions MAY implement
+Compatible version negotiation if it prefers one version over another.
```
DavidSchinazi, line 408: similarly, add reference

draft-ietf-quic-version-negotiation.md
```diff
@@ -387,6 +387,40 @@ lifetime of the connection; endpoints MUST NOT change the version that they
...
+
+If a client sends a Version Information Transport Parameter in a packet using a
+version that is compatible with a version advertised in the transport parameter,
+it MUST support compatible version negotiation.
+
+Therefore, clients can only avoid support for compatible negotiation by either
+(1) never attempting connection based on a Version Negotiation packet, and
+electing not to include the transport parameter; or (2) never including a
+supported version in that transport parameter with which the one in use is
+compatible.
```
DavidSchinazi, line 423:
```suggestion
compatible version in the transport parameter.
```

draft-ietf-quic-version-negotiation.md
```diff
@@ -387,6 +387,40 @@ lifetime of the connection; endpoints MUST NOT change the version that they
...
+Hosts can support this standard incrementally.
+
+{{INV}} and {{QUIC}} already describe how servers send Version Negotiation
+packets in response to long header packets with unsupported version fields, even
+if they support only one version. Doing so prevents long timeouts at the client,
+but servers might not send the packet multiple times for the same connection, or
+when under stress.
+
+If a server supports two or more QUIC versions and sends Version Negotiation
+packets, it MUST send the Version Information Transport Parameter {{vers-info}},
```
DavidSchinazi, line 401: I feel like this just repeats the requirements from the downgrade section. Do we need it?

draft-ietf-quic-version-negotiation.md
```diff
@@ -387,6 +387,40 @@ lifetime of the connection; endpoints MUST NOT change the version that they
...
+Negotiation packet with a connection attempt using one of the advertised
+versions, MUST include a Version Information Transport Parameter and process the
+server's transport parameter in accordance with {{downgrade}}.
+
+If a client sends a Version Information Transport Parameter in a packet using a
+version that is compatible with a version advertised in the transport parameter,
+it MUST support compatible version negotiation.
+
+Therefore, clients can only avoid support for compatible negotiation by either
+(1) never attempting connection based on a Version Negotiation packet, and
```
DavidSchinazi, line 420: nit: use bullets

draft-ietf-quic-version-negotiation.md
```diff
@@ -387,6 +387,40 @@ lifetime of the connection; endpoints MUST NOT change the version that they
 consider to be in use based on the Version field of long headers as that field
 could be forged by attackers.
 
+# Minimum support
+
+Hosts can support this standard incrementally.
```
DavidSchinazi, line 392: I think we should expand on this to explain why it might be desirable to only partially support this spec

draft-ietf-quic-version-negotiation.md
martinduke: Struggled with the commutative property here, thanks.

martinduke:
i'm not 100% sure what you wanted in some cases, but I addressed all your comments.

draft-ietf-quic-version-negotiation.md
```diff
@@ -387,6 +387,41 @@ lifetime of the connection; endpoints MUST NOT change the version that they
...
+
+If a server supports two or more QUIC versions and sends Version Negotiation
+packets, it MUST follow the requirements in {{vers-info}} and {downgrade}}.
+Failure to comply would leave the connection vulnerable to downgrade attacks, or
+cause connection failure, if the client initiates Incompatible Version
+Negotiation ({{incompat-vn}}).
+
+A server that supports two or more compatible QUIC versions MAY implement
+Compatible version negotiation ({{compat-vn}}) if it prefers one version over
+another.
```
ekr, line 409: I would rewrite this almost entirely non-normatively by describing what is technically possible.

E.g.,

It is possible to implement only VN or compatible version negotiation. Servers which implement only compatible version negotiation will not be able to interoperate with clients which send compatible Initial values. Servers which implement only VN will not be able to select between multiple acceptable versions [I think this is true.]

draft-ietf-quic-version-negotiation.md
```diff
@@ -387,6 +387,41 @@ lifetime of the connection; endpoints MUST NOT change the version that they
...
+
+A client that supports two or more versions, and chooses to respond to a Version
+Negotiation packet with a connection attempt using one of the advertised
+versions, MUST include a Version Information Transport Parameter and process the
+server's transport parameter in accordance with {{downgrade}}.
+
+If a client sends a Version Information Transport Parameter in a packet using a
+version that is compatible with a version advertised in the transport parameter,
+it MUST support compatible version negotiation.
```
ekr, line 418: Isn't this stuff required elsewhere.

draft-ietf-quic-version-negotiation.md
```diff
@@ -387,6 +387,41 @@ lifetime of the connection; endpoints MUST NOT change the version that they
...
+# Minimum support
+
+Hosts can support this standard incrementally to simplify implementation, if
+not all features of the design are desired.
+
+{{INV}} and {{QUIC}} already describe how servers send Version Negotiation
+packets in response to long header packets with unsupported version fields, even
+if they support only one version. Doing so prevents long timeouts at the client,
+but servers might not send the packet multiple times for the same connection, or
+when under stress.
```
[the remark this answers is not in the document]
martinduke: It's a summary of Sec 6.1 of RFC9000. The usual case would be an INITIAL with 0RTT packets -- if the server can recognize the packet type, it can limit the number of packets to 1 statelessly.  But one can certainly imagine some sort of global state saying "I am under DoS attack, stop sending VN". 

The important thing this paragraph is saying is reminding the reader that other RFCs say that you might get a VN packet, but you might not. I'm happy to get a suggested rewording.

draft-ietf-quic-version-negotiation.md
martinduke: OK

draft-ietf-quic-version-negotiation.md
martinduke: If so, I can't find it, but if one reads and fully understands the doc it's an unavoidable conclusion.I thought it would be useful to make it explicit

draft-ietf-quic-version-negotiation.md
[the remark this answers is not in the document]
ekr: Well, the server is free to not respond to *any* message it wants. What's unique about VN.

draft-ietf-quic-version-negotiation.md
ekr: OK, I guess for now it can go here, but I think generally it should go elsewhere.

martinduke:
Closing  -- authors preferred #83 to this one

PR #83, opened 2022-03-07
min support, but much much shorter

### LINKED ISSUE #80: Differing levels of support for the standard

martinduke:
Fixes #80 

After pushback from @ekr, I think this covers the main point with a whole lot less text.

### DIFF
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -287,8 +287,8 @@ remain stable for the entire handshake (as QUIC version 1 does), then this
 applies to the entire handshake, including the first flight.
 
 Note also that the client can disable compatible version negotiation by
-omitting the Version Information Transport Parameter, or not reporting other
-versions in it.
+only including the Chosen Version in the Other Versions field of the Version
+Information Transport Parameter.
 
 # Version Information {#vers-info}
 
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -286,6 +286,9 @@ version. For instance, if the negotiated version requires that the 5-tuple
 remain stable for the entire handshake (as QUIC version 1 does), then this
 applies to the entire handshake, including the first flight.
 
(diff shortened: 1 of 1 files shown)


ekr (approved):
LGTM

draft-ietf-quic-version-negotiation.md
```diff
@@ -286,6 +286,9 @@ version. For instance, if the negotiated version requires that the 5-tuple
 remain stable for the entire handshake (as QUIC version 1 does), then this
 applies to the entire handshake, including the first flight.
 
+Note also that the client can disable compatible version negotiation by
+omitting the Version Information Transport Parameter, or not reporting other
+versions in it.
```
DavidSchinazi, line 291: Omitting the transport parameter removes the version from the TLS transcript which weakens security. I would instead:
```suggestion
only including the Chosen Version in the Other Versions field of the Version
Information Transport Parameter.
```

DavidSchinazi (approved):

PR #85, opened 2022-03-07
clarify handling of multiple ALPNs

### LINKED ISSUE #81: Don't assume a single protocol
The document currently says:

> When a client creates a QUIC connection, its goal is to use an application layer protocol. Therefore, when considering which versions are compatible, clients will only consider versions that support the intended application layer protocol.

However, as we see with HTTP over TCP, clients routinely offer "h2", "http/1.1" and are content with either selection.  Particularly if ALPN tokens map 1:1 to QUIC versions as specified in H3, the client's goal might be to use any of a collection of application-layer protocols.

Perhaps tweak this to "...support one of the intended application layer protocols."?  The first statement might or might not need to be expanded as well.

martinduke:
Fixes #81

Mix recommendations of @MikeBishop and @DavidSchinazi

### DIFF
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -189,8 +189,11 @@ endpoint might be aware of the compatibility document while the other may not.
 
 When a client creates a QUIC connection, its goal is to use an application layer
 protocol. Therefore, when considering which versions are compatible, clients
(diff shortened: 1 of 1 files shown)


ekr:
LGTM

DavidSchinazi (approved):

PR #86, opened 2022-03-07
handle TP with version = 0

### LINKED ISSUE #76: A value of 0 in the transport parameter
Should an endpoint be required to reject a transport parameter that contains 0?

martinduke:
Fixes #76.

### DIFF
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -340,6 +340,11 @@ in their Other Versions list. Those versions are reserved to exercise version
 negotiation (see the Versions section of {{QUIC}}), and will never be selected
 when choosing a version to use.
 
+If a client receives this parameter with any version field as zero, it MUST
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-version-negotiation.md
```diff
@@ -340,6 +340,11 @@ in their Other Versions list. Those versions are reserved to exercise version
 negotiation (see the Versions section of {{QUIC}}), and will never be selected
 when choosing a version to use.
 
+If a client receives this parameter with any version field as zero, it MUST
```
DavidSchinazi, line 343: I don't like this difference between client and server. How about instead you add this to paragraph below that starts with `Both endpoints MUST parse...`:

`If an endpoint receives a Chosen Version equal to zero, or any Other Version equal to zero, it MUST treat it as a parse error.`

DavidSchinazi (approved):

PR #87, opened 2022-03-07
Clarify negotiated version

### LINKED ISSUE #67: Specify impact on encryption keys
I am reading the draft, and I have doubts on how this is supposed to affect encryption keys.

Suppose a classic exchange: the client sends a couple of initial packets, the server parses them and acknowledges them, the server then sends initial and handshake packets. I think we have the following constraints:

1) The server will only complete the compatible version negotiation after receiving "enough" of the client's initial packets, i.e., enough to parse the QUIC transport parameters.
2) The server transport parameters are sent as encrypted extensions. The client will only receive them after receiving "enough" handshake packets.
3) The client could learn that the server has chosen a new version at any point during the handshake, by receiving a new version field in a long header packet --  Initial or handshake.

So we get possible variations:

1) Server receives client first flight with V1, may reply with V1 ACK until the flight has been received, but uses V2 for the reminder of the handshake, i.e., sends server hello in a V2 Initial packet, sends the encrypted parameters and the reminder of the server first flight in V2 handshake packets. Upon reception and successful decryption of initial V2 packet, the client switches to V2.

2) Same but server sends all Initial packets using V1, all handshake packets using V2. Client switches to V2 upon reception and successful decryption of V2 handshake packet.

3) Server keeps using V1 for initial and handshake packets, switches to V2 for 1RTT packets. Client keeps using V1 until the handshake completes, switches to V2 for 1RTT packets.

Variation 3 is the easiest to implement, but I suspect that "the server converts the client's first flight to that version and replies to the client as if it had received the converted first flight" actually implies variation 1. That variation requires that the client makes a pretty expensive decision, recomputing the encryption keys using the new version's parameters upon receiving a packet with a "negotiated" version in the version header. Since the encryption key of initial packets is well known, this is yet another avenue for third parties to send poisoned packets and cause the client to fail the connection, but I suppose that will not stop us because there are many such attacks already.

In any case, the current text is not very explicit. I think that we will save a lot of interop issues by being more explicit, maybe presenting the expected flow of packets.

### LINKED ISSUE #72: validating Version Information on compatible negotiation
I'm reading editor's draft.

It says that both peers MUST parse Version Information. This is good.

However, a client MUST validate Version Information only on incompatible version negotiation ("acted on a Version Negotiation packet"). I think the same validation should be done on compatible version negotiation.

### LINKED ISSUE #79: What is the negotiated version during the handshake?
The [editor's copy of the v2 draft](https://quicwg.org/quic-v2/draft-ietf-quic-v2.html#name-version-negotiation-conside) currently states:
> Both endpoints MUST send Handshake or 1-RTT packets using the negotiated version. An endpoint MUST drop packets using any other version.

Let's assume the following scenario:
* the server supports both v1 and v2, and is configured to perform compatible negotiation from 1 to 2 when possible
* client sends a QUIC v1 initial with "Other Versions" = (1, 2)
* the server decides to switch to v2, and sends its initial ServerHello as v2, followed by a v2 handshake packet with the first half of the certificate - the transport parameters will follow in a later packet
* when the client receives the v2 handshake packet, it technically doesn't yet know the "negotiated version" because it hasn't yet parsed the transport parameters

Conceptually, this sounds like a terminology issue with the VN draft: at that point in time the client does know that the server wants to use v2, because it's parsed a v2 ServerHello - but I think we need a name for that in the VN draft so that the v2 draft can refer to it for the MUST above.

cc @martinduke @martinthomson -- the more I think about it the more I think that [MT's suggestion](https://github.com/quicwg/version-negotiation/pull/42#discussion_r630689250) to "use instead of validate" is making the spec and implementation writing a lot harder - the idea makes sense in theory, but in practice it complicates the handshake. Maybe we go back to something like "the negotiated version is the one seen in long headers, and the Chosen Version in the TP exists to validate that it matches after the fact"

DavidSchinazi:
Fixes #67 
Fixes #72 
Fixes #79

### DIFF
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -280,9 +280,13 @@ server in its reply is refered to as the "negotiated version". The server MUST
 NOT reply with a version that is not present in the client's compatible
 versions, unless it is the original version.
 
-If the server does not find a compatible version, it will use the original
-version if it supports it, and if it doesn't then the server will perform
-incompatible version negotiation instead, see {{incompat-vn}}.
+Clients will be made aware of compatible version negotiation by seeing a change
+in the QUIC long header Version field. It is possible for the server to
+initially send packets with the original version before switching to the
+negotiated version (for example, this can happen when the client's Version
+Information structured spans multiple packets; in that case the server might
+acknowledge the first packet in the original version and later switch to a
+different negotiated version).
 
 Note that, after the first flight is converted to the negotiated version, the
 handshake completes in the negotiated version. The entire handshake (including
@@ -294,6 +298,10 @@ applies to the entire handshake, including the first flight.
 Note also that the client can disable compatible version negotiation by
 only including the Chosen Version in the Other Versions field of the Version
 Information Transport Parameter.
+
+If the server does not find a compatible version, it will use the original
+version if it supports it, and if it doesn't then the server will perform
+incompatible version negotiation instead, see {{incompat-vn}}.
 
 # Version Information {#vers-info}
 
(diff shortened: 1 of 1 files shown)


martinduke (approved):
LGTM

PR #102, opened 2022-04-05
Implementaion leeway for 2 connections

### LINKED ISSUE #95: Is incompatible VN one or two connections?
I am not sure that I agree with this as some state carries across from the first connection attempt to the second (and hopefully successfully completed) connection.

(Also, I've implemented this within the same connection object.  I know that other implementations differ here, so what I'm looking for is a little finesse in how this is framed.)

DavidSchinazi:
Fixes #95.

### DIFF
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -233,7 +233,7 @@ version that follows.
 
 Note that this separation across two connections is conceptual: it applies to
 normative requirements on QUIC connections, but does not require implementations
-to use two distinct connection objects.
+to internally use two distinct connection objects.
 
 
 ## Client Choice of Original Version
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-version-negotiation.md
```diff
@@ -231,6 +231,10 @@ across two QUIC connections: the connection attempt prior to receiving the
 Version Negotiation packet is distinct from the connection with the incompatible
 version that follows.
 
+Note that this separation across two connections is conceptual: it applies to
+normative requirements on QUIC connections, but does not require implementations
+to use two distinct connection objects.
```
ekr, line 236:
```suggestion
to internally use two distinct connection objects.
```

ekr (approved):

PR #103, opened 2022-04-05
Clarify client choice

### LINKED ISSUE #96: Choosing the original version
Section 2.5 says unconditionally that the client chooses the most compatible version.

This isn't right, because the most compatible version isn't what it is looking to end up on.  What the client should choose is probably the most compatible version of those that are compatible (directionality ?) with its most preferred version.

DavidSchinazi:
Fixes #96

### DIFF
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -236,11 +236,11 @@ version that follows.
 
 When the client picks its original version, it will try to avoid incompatible
 version negotiation to save a round trip. Therefore, the client SHOULD pick an
-original version to maximize the combined probability of both:
+original version to maximize the combined probability that both:
 
 * The server knows how to parse first flights from the original version.
 
-* The original version is compatible with the client's prefered version.
+* The original version is compatible with the client's preferred version.
 
 Without additional information, this could mean selecting the oldest version
 that the client supports.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-version-negotiation.md
```diff
@@ -234,11 +234,16 @@ version that follows.
...
 ## Client Choice of Original Version
 
-The client's first connection attempt SHOULD be made using the version that the
-server is most likely to support. The client selects the version most likely to
-be supported from the versions that are compatible with the client's most
-preferred version. Without additional information this could mean selecting the
-oldest version that the client supports.
+When the client picks its original version, it will try to avoid incompatible
+version negotiation to save a round trip. Therefore, the client SHOULD pick an
+original version to maximize the combined probability of both:
```
ekr, line 239:
```suggestion
original version to maximize the combined probability that both:
```

draft-ietf-quic-version-negotiation.md
```diff
@@ -234,11 +234,16 @@ version that follows.
...
+
+* The server knows how to parse first flights from the original version.
+
+* The original version is compatible with the client's prefered version.
```
ekr, line 243:
```suggestion
* The original version is compatible with the client's preferred version.
```

ekr (approved):
LGTM with the following edits.

PR #104, opened 2022-04-05
Clarify compatible

### LINKED ISSUE #90: What does "compatible" VN mean
Consider the case where the client and server both support v1 and v2 and prefer v2.

In the current draft, if the client puts v1 in the long header and the server responds with v2, that's called "compatible version negotiation". OTOH, if the client puts v2 in the long header and the server responds with v2, that's called, well, I'm not quite sure what. I guess not negotiation?

I think a better way to think about this is that the client is sending an Initial that is compatible with both versions and so both of these cases are compatible version negotiation.

### LINKED ISSUE #94: Original version in compatible negotiation
> The server MUST NOT reply with a version that is not present in the client's compatible versions, unless it is the original version.

Original version was defined as the version a client attempts prior to a VN packet.  If there is a VN packet, I think that it is still possible to do a compatible upgrade from the version in the client's second attempt.  In that case, "original version" isn't the right term to use when talking about compatible upgrades.

ekr:
Fixes #90 
Fixes #94

### DIFF
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -558,16 +558,6 @@ negotiation to work with servers that only support QUIC version 1. Note that
 implementations which wish to use version negotiation to negotiate versions
 other than QUIC version 1 will need to implement this draft.
 
-## Special Case Behavior for QUIC Version 1
-
-QUIC version 1 was standardized prior to the publication of this
-document and therefore QUICv1 servers may not support this
-extension. If a client implementations whch start a QUICv1 connection
-in response to a version negotiation packet and the server omits,
-the 
-
-
-
 
 # Security Considerations
 
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -78,7 +78,7 @@ It is beneficial to avoid additional round trips whenever possible, especially
 given that most incremental versions are broadly similar to the the previous
 version. This specification also defines a simple version negotiation mechanism
 which leverages similarities between versions and can negotiate between the set
-of "compatible" versions jointly supported by without additional round trips.
+of "compatible" versions without additional round trips.
 
 
 ## Conventions and Definitions
@@ -111,16 +111,16 @@ Upon receiving this first flight, the server verifies whether it knows how to
 parse first flights from the original version. If it does not, then it starts
 incompatible version negotiation, see {{incompat-vn}}, which causes the
 client to initiate a new connection with a different version.
-For instance, if the client initiates a connection with version 1
+For instance, if the client initiates a connection with version A
 and the server starts incompatible version negotiation and the client
-then initiates a new connection with version 2, we say that
-the first connection's client chosen version is 1, the second connection's
-client chosen is 2, and the original version for the entire
-sequence is 1.
+then initiates a new connection with version B, we say that
+the first connection's client chosen version is A, the second connection's
+client chosen version is B, and the original version for the entire
+sequence is A.
 
 If the server can parse the first flight, it can either establish the
 connection using the client's chosen version, or it MAY select any other
-compatible version, as described in {{compat-vn}}. 
+compatible version, as described in {{compat-vn}}.
 
 Note that it is possible for a server to have the ability to parse the first
 flight of a given version without fully supporting it, in the sense that it
@@ -199,8 +199,8 @@ chosen version, it can extract the client's Version Information structure
 client knows its first flight is compatible with.
 
 In order to perform compatible version negotiation, the server MUST select
-one of these versions that (1) it supports and (2) it knows is
-compatible with the client's chosen version. Once the server has
+one of these versions that (1) it supports and (2) it knows
+the client's chosen version to be compatible with. Once the server has
 selected a version, termed the "negotiated version", it then attempts
 to convert the client's first flight into that version, and replies to
 the client as if it had received the converted first flight.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-version-negotiation.md
```diff
@@ -76,7 +78,7 @@ It is beneficial to avoid additional round trips whenever possible, especially
 given that most incremental versions are broadly similar to the the previous
 version. This specification also defines a simple version negotiation mechanism
 which leverages similarities between versions and can negotiate between the set
-of "compatible" versions without additional round trips.
+of "compatible" versions jointly supported by without additional round trips.
```
DavidSchinazi, line 81: supported by ... ?

draft-ietf-quic-version-negotiation.md
```diff
@@ -95,18 +97,30 @@ This document specifies two means of performing version negotiation: one
...
-downgrade attacks (see {{downgrade}}).
+The client initiates a QUIC connection by choosing an initial version
+and sending a first flight of QUIC packets with a long header to the
+server {{INV}}.  The client's first flight includes Version
+Information (see {{vers-info}}) which will be used to optionally
+enable compatible version negotation (see {{compat-vn}}), and to
+prevent version downgrade attacks (see {{downgrade}}).  We'll refer to
+the version of the very first packets the client sends as the "original version"
+and the version of the first packets the client sends in a
+a given QUIC connection as the "client's chosen version".
```
DavidSchinazi, line 108:
```suggestion
given QUIC connection as the "client's chosen version".
```

draft-ietf-quic-version-negotiation.md
```diff
@@ -95,18 +97,30 @@ This document specifies two means of performing version negotiation: one
...
 
 Upon receiving this first flight, the server verifies whether it knows how to
 parse first flights from the original version. If it does not, then it starts
-incompatible version negotiation, see {{incompat-vn}}. If the server can parse
-the first flight, it can either establish the connection using the original
-version, or it MAY attempt compatible version negotiation, see {{compat-vn}}.
+incompatible version negotiation, see {{incompat-vn}}, which causes the
+client to initiate new connection with a different version.
```
DavidSchinazi, line 113:
```suggestion
client to initiate a new connection with a different version.
```

draft-ietf-quic-version-negotiation.md
```diff
@@ -95,18 +97,30 @@ This document specifies two means of performing version negotiation: one
...
+For instance, if the client initiates a connection with version 1
```
DavidSchinazi, line 114: can we say A and B instead of 1 and 2?

draft-ietf-quic-version-negotiation.md
```diff
@@ -178,27 +193,32 @@ endpoint might be aware of the compatibility document while the other may not.
...
-
-Clients will be made aware of compatible version negotiation by seeing a change
-in the QUIC long header Version field. It is possible for the server to
-initially send packets with the original version before switching to the
+When the server can parse the client's first flight using the client's
+chosen version, it can extract the client's Version Information structure
+(see {{vers-info}}). This contains the list of versions that the
+client knows its first flight is compatible with.
+
+In order to do compatible version negotiation, the server MUST select
```
DavidSchinazi, line 201:
```suggestion
In order to perform compatible version negotiation, the server MUST select
```

draft-ietf-quic-version-negotiation.md
```diff
@@ -178,27 +193,32 @@ endpoint might be aware of the compatibility document while the other may not.
...
+one of these versions that (1) it supports and (2) it knows is
+compatible with the client's chosen version. Once the server has
```
DavidSchinazi, line 203: this is backwards. the client's chosen version needs to be compatible with the selected version not the other way around

draft-ietf-quic-version-negotiation.md
```diff
@@ -178,27 +193,32 @@ endpoint might be aware of the compatibility document while the other may not.
...
+selected a version, termed the "negotiated version", it then attempts
+to convert the client's first flight into that version, and replies to
+the client as if it had received the converted first flight.
+
+If those formats are identical, as in cases where the negotiated
+version is the same as the client's chosen version, then this will be
+the identity transform). If the first flight is correctly formatted,
```
DavidSchinazi, line 210:
```suggestion
the identity transform. If the first flight is correctly formatted,
```

draft-ietf-quic-version-negotiation.md
```diff
@@ -178,27 +193,32 @@ endpoint might be aware of the compatibility document while the other may not.
...
+then this conversion process cannot fail by definition of the first
+flight being compatible; if the server is unable to convert the first
+flight, it MUST respond with CONNECTION_CLOSE and an error of
+PROTOCOL_VIOLATION.
```
DavidSchinazi, line 214: That's QUICv1-specific. Let's just say abort the handshake

DavidSchinazi (commented):
Looks good

DavidSchinazi (approved):

draft-ietf-quic-version-negotiation.md
```diff
@@ -212,17 +231,36 @@ Note also that the client can disable compatible version negotiation by
...
+scenario may occur:
+
+~~~
+    Client                                          Server
+
+    Chosen = A, Other Versions = (A, B) ----------------->
+    <------------------------ Version Negotiation = (D, C)
+
+    Chosen = C, Other Versions = (C, D) ----------------->
+    <-------------------------------------- Negotiated = D
```
martinthomson, line 250: Want to show Other Versions here too?

Fun fact: that list could include B and E.

DavidSchinazi: I feel like adding this could add more confusion than it'd help remove?

martinthomson: Oh yes, for certain, but you might at least show that there are other versions here, so maybe "Other Versions = (...)".

DavidSchinazi: Good point, wrote up PR #105

PR #106, opened 2022-04-22
Editorial clarification on advertising newer versions

mirjak:


draft-ietf-quic-version-negotiation.md
```diff
@@ -282,7 +282,8 @@ original version to maximize the combined probability that both:
 * The original version is compatible with the client's preferred version.
 
 Without additional information, this could mean selecting the oldest version
-that the client supports.
+that the client supports, while adversiting newer compatible versions in the
```
ekr, line 285:
```suggestion
that the client supports, while advertising newer compatible versions in the
```

PR #107, opened 2022-04-22
Refactor to create Definitions section

DavidSchinazi:


draft-ietf-quic-version-negotiation.md
```diff
@@ -80,13 +80,33 @@ which leverages similarities between versions and can negotiate between the set
...
+
+* In the context of a given QUIC connection, the "first flight" of packets
+  refers to the set of packets the client creates and sends to initiate the
+  connection before it has heard back from the server.
+
+* In the context of a given QUIC connection, the "client's chosen version" is
+  the QUIC version of the connection's first flight.
+
+* The "original version" is the QUIC version of the very first packet the client
+  sends to the server. If version negotiation spans multiple connections, the
```
nibanks, line 100: > spans multiple connections

Are you referring to the different initial flights between incompatible version negotiation? Do we really consider that multiple connections? I don't think it's ever been referred it this way before.

I think I understand what each of these terms mean, or at least what you're trying to define them as, but how they are worded here actually confuses me more.

DavidSchinazi: > Are you referring to the different initial flights between incompatible version negotiation? Do we really consider that multiple connections? I don't think it's ever been referred it this way before.

Yes, as described in the "Connections and Version Negotiation" section.

> I think I understand what each of these terms mean, or at least what you're trying to define them as, but how they are worded here actually confuses me more.

Can you help explain what you find confusing? It's not very actionable for us editors otherwise.

nibanks: AFAIU, there are not multiple connections involved here. The client is making one connection, that is going through the version negotiation process. I don't consider incompatible negotiation a new connection. When you say first flight, I believe you mean the very first flight, and then also when the client responds for an incompatible negotiation; but you that doesn't align with my understanding of this text: "before it has heard back from the server". If you're doing incompatible negotiation, that's after you've heard from the server.

DavidSchinazi: > AFAIU, there are not multiple connections involved here. The client is making one connection, that is going through the version negotiation process. I don't consider incompatible negotiation a new connection.

If you have thoughts on the "Connections and Version Negotiation" section, please take them to an issue instead of a comment thread on an editorial PR.

> When you say first flight, I believe you mean the very first flight, and then also when the client responds for an incompatible negotiation; but you that doesn't align with my understanding of this text: "before it has heard back from the server". If you're doing incompatible negotiation, that's after you've heard from the server.

I think that is covered by "In the context of a given QUIC connection". Am I missing something?

mirjak (approved):

draft-ietf-quic-version-negotiation.md
```diff
@@ -80,13 +80,33 @@ which leverages similarities between versions and can negotiate between the set
...
+  original version is equal to the client's chosen version of the first QUIC
+  connection.
+
+* The "negotiated version" is the QUIC version in use on the connection once the
+  version negotiation process completes.
+
+* The "Maximum Segment Lifetime" (MSL) represents the time a QUIC packet can
+  exist in the network. Implementations can make this configurable, and a
+  RECOMMENDED value is one minute.
```
martinthomson, line 109: This is a little weird in that "segment" is a TCP concept.  You might need to acknowledge the origin of the term in order to explain that.

draft-ietf-quic-version-negotiation.md
martinthomson: A forward reference to the bit about connections and incompatible negotiation might help address some of this confusion.

martinthomson (approved):

draft-ietf-quic-version-negotiation.md
DavidSchinazi: Good point, added

draft-ietf-quic-version-negotiation.md
```diff
@@ -96,14 +117,11 @@ This document specifies two means of performing version negotiation: one
 and one "compatible" that allows saving the round trip but only applies when the
 versions are compatible.
 
-The client initiates a QUIC connection by choosing an initial version and
+The client initiates a QUIC connection by choosing an original version and
```
ekr, line 120: This should still say "initial" because incompatible VN is two different connections.

ekr (changes requested):

draft-ietf-quic-version-negotiation.md
DavidSchinazi: But this sentence only refers to the first connection. Maybe we need a better name that refers to the set of related connections.

ekr: I don't get that out of this text. 

The client initiates a QUIC connection. It gets a VN. It then initiates a new QUIC connection.

DavidSchinazi: That's only true of the first connection, see Section 4:
> A client that makes a connection attempt based on information received from a Version Negotiation packet MUST ignore any Version Negotiation packets it receives in response to that connection attempt.

ekr: That statement is true, but irrelevant to this sentence.

DavidSchinazi: Sure, I don't feel too strongly
```suggestion
The client initiates a QUIC connection by choosing an initial version and
```

ekr:
LGTM

PR #108, opened 2022-04-22
Clarify when long header version can mismatch

DavidSchinazi:


nibanks (approved):

mirjak (approved):

draft-ietf-quic-version-negotiation.md
```diff
@@ -311,7 +311,8 @@ The content of each field is described below:
 Chosen Version:
 : The version that the sender has chosen to use for this connection. In most
 cases, this field will be equal to the value of the Version field in the long
-header that carries this data.
+header that carries this data; however future versions or extensions can choose
+to set different values in the long header Version field.
```
ekr, line 315: I think this needs some work, because long header Version field *indicates* the version.

Perhaps rewrite the whole thing.

"The version that the sender has chosen to use for this connection as indicated by the value of the Version field in the long header that carries this data. In most cases, these values will be the same, but future versions of QUIC might opt to mask or otherwise change the version field that appears on the wire."

ekr (changes requested):

draft-ietf-quic-version-negotiation.md
DavidSchinazi: That feels more confusing to me. The value of the Version field is what's sent on the wire.

ekr: Well, then how can it be different from what's in this field?

DavidSchinazi: Here's an example of a way we could have designed the v1-to-v2 compatible upgrade:
* client sends INITIAL as `long_header_version=v1` with `chosen_version=v1, other_versions=(v1,v2)`
* server sends INITIAL and HANDSHAKE as `long_header_version=v1` with `chosen_version=v2`
* clients sends HANDSHAKE as `long_header_version=v1`
* then all 1-RTT packets use v2

That's not how v2 looks today, but that was a possible design. In that scenario, the long_header_version doesn't match the chosen_version

ekr: I think this document should specify that that that's not how this works.

ekr: Indeed, we effectively do:
> Clients can determine the server's negotiated version by examining
the QUIC long header Version field. It is possible for the server to
initially send packets with the client's chosen version before switching to the
negotiated version (for example, this can happen when the client's Version
Information structured spans multiple packets; in that case the server might
acknowledge the first packet in the client's chosen version and later switch to a
different negotiated version).

DavidSchinazi: That text doesn't preclude sending the server's transport parameters before switching to the negotiated version

ekr (approved):

PR #109, opened 2022-06-03
Leave how to determine versions up to versions

ekr:


draft-ietf-quic-version-negotiation.md
```diff
@@ -227,12 +227,16 @@ If the first flight is correctly formatted, then this conversion process cannot
 fail by definition of the first flight being compatible; if the server is unable
 to convert the first flight, it MUST abort the handshake.
 
-Clients can determine the server's negotiated version by examining the QUIC long
-header Version field. It is possible for the server to initially send packets
-with the client's chosen version before switching to the negotiated version (for
-example, this can happen when the client's Version Information structure spans
-multiple packets; in that case the server might acknowledge the first packet in
-the client's chosen version and later switch to a different negotiated version).
+Any version of QUIC MUST specify how to determine whether it is has
```
DavidSchinazi, line 230: Here's a proposal on a slightly different way to frame this:

```
If a document specifies that a QUIC version is compatible with another,
that document MUST specify the mechanism by which endpoints are
made aware of the negotiated version. An example of such a mechanism
is to have the client determine the server's negotiated version by
examining the QUIC long header Version field. Note that, in this example
mechanism, it is possible for the server to initially send packets
with the client's chosen version before switching to the negotiated version
(this can happen when the client's Version Information structure spans
multiple packets; in that case the server might acknowledge the first packet in
the client's chosen version and later switch to a different negotiated version).
Any set of mutually compatible versions SHOULD use the same mechanism.
```

ekr: I can live with this modulo "by which endpoints" -> "by which clients".

PR #110, opened 2022-06-29
VN/v2 shepherd writeup

mjoras:


writeups/vn.txt
```diff
@@ -0,0 +1,133 @@
...
+   final version of the document written in a formal language, such as XML code,
+   BNF rules, MIB definitions, CBOR's CDDL, etc.
+Normal linting.
+
+### Document Shepherd Checks
+
+9. Based on the shepherd's review of the document, is it their opinion that this
+   document is needed, clearly written, complete, correctly designed, and ready
+   to be handed off to the responsible Area Director?
+I believe these documents are complete and clear and ready for AD review.
```
DavidSchinazi, line 64: Maybe also mention "needed" and "correctly designed"?

writeups/vn.txt
```diff
@@ -0,0 +1,133 @@
...
+10. Several IETF Areas have assembled [lists of common issues that their
+    reviewers encounter][6]. For which areas have such issues been identified and addressed?
+    For which does this still need to happen in subsequent reviews?
+N/A.
+
+11. What type of RFC publication is being requested on the IETF stream ([Best
+    Current Practice][12], [Proposed Standard, Internet Standard][13],
+    [Informational, Experimental or Historic][14])? Why is this the proper type
+    of RFC? Do all Datatracker state attributes correctly reflect this intent?
+Proposed Standard.
```
DavidSchinazi, line 75: The question also asks Why?

DavidSchinazi (approved):
Looks good modulo two comments

writeups/vn.txt
LPardue: The appropriate answer is "Because". (Snarky, because the question is a bit dumb)

DavidSchinazi: I don't disagree with the question being dumb, but if you don't answer it here it might lead to a DISCUSS so an overzealous AD can get it answered

LPardue: Note that this writeup is based on the WIP shepherd doc PR that addresses some of the community feedback levied towards the latest writeup QA template. The usefulness of this question is an unresolved thread as far as I'm concerned - see https://github.com/ietf-tools/datatracker/pull/4042#discussion_r888515151

DavidSchinazi: Fair, but saying "Because the WG reached consensus on this being the long-term VN mechanism for QUIC" doesn't hurt and might save the editors some time later :)

LPardue: sure, we aren't working to any word limit, it still boils down to "because the WG declared consensus to do so". And if that's ok, we can just make it a stock answer, which seems to defeat the purpose of the question as its currently posed.

DavidSchinazi: To be clear I'm fully in agreement with working to improve the template, and would like to see that question removed. I'm just asking for us to not be guinea pigs with this document, I don't want to have discussions about the merit of the template question during this draft's IESG review

LPardue: That's a fair concern that I take on board. We don't want to make our editors lives hard for the sake of it, nor tie up the progress of this document. 

But I'd honestly be surprised if someone objected to the statement "because we have WG consensus it is" while being happy with "because we have WG consensus that the draft $foo explains how to do $foo is appropriate"

DavidSchinazi: Oh the only difference is that it sounds slightly less snarky :-)

writeups/vn.txt
LPardue: yeah agree here, the answer is an easy direct one - omission here sticks out

writeups/vn.txt
```diff
@@ -0,0 +1,133 @@
...
+No.
+
+20. Describe the document shepherd's review of the IANA considerations section,
+    especially with regard to its consistency with the body of the document.
+    Confirm that all aspects of the document requiring IANA assignments are
+    associated with the appropriate reservations in IANA registries. Confirm
+    that any referenced IANA registries have been clearly identified. Confirm
+    that each newly created IANA registry specifies its initial contents,
+    allocations procedures, and a reasonable name (see [RFC 8126][11]).
+The V2 document references the QUIC version registry correctly.
```
LPardue, line 128: VN has IANA considerations too, so we should cover that, possibly noting that the intended codepoint is in the 0-63 range, which is Specification Required

PR #122, opened 2022-09-14
Clarify validations on converted first flight

### LINKED ISSUE #118: AD review : unclear reference to "first flight"
https://github.com/quicwg/version-negotiation/blob/e193e1264887ffab5919d529ea21c0023989c1ec/draft-ietf-quic-version-negotiation.md?plain=1#L249

here it is not clear to me if we are talking about first flight or converted first flight. I am assuming the negotiated version is in the converted first flight. May be we can be more clear here.

DavidSchinazi:
Fixes #118


zaheduzzaman:
thanks for the fix.

PR #123, opened 2022-09-14
Make incompatible VN more normative

### LINKED ISSUE #119: AD review : When does the negotiation end?
https://github.com/quicwg/version-negotiation/blob/e193e1264887ffab5919d529ea21c0023989c1ec/draft-ietf-quic-version-negotiation.md?plain=1#L275

in this example, it would be very stupid of the client to select version C and start another negotiation after the sever has indicated D is what it prefers. The server can again select D as preferred and there exits a potential loop. However, it is possible. What are the rule we have in the specification that stops this kind of loop?

DavidSchinazi:
Fixes #119


zaheduzzaman (commented):
LGTM

PR #124, opened 2022-09-26
Updates RFC 8999

### LINKED ISSUE #115: AD Review : Should this specification add a update tag?
I would like to have a discussion to determine whether this specification should update RFC9000 or not.

This version negotiation draft adds new functionality on top of RFC9000 using the version negotiation packet from RFC8999. The update tag might be seen as useful to let other aware of the added mechanism for version negotiation, also because it actually changes how section 6 of RFC9000 deals with QUIC versions, which mentions potential future work regarding the negotiation. On the other hand, this new mechanism is not really a must to implement to make things works for RFC9000and both mechanism can co-exists, hence falls into QUIC extension categories. It seems all about how we want the future QUIC versions to get deployed.

I have checked with WG chairs and authors that we didn't really discussed this in the working group. I think having a discussion now would be useful to know where the WG stands on this and have an response ready for potential questions (from reviews) along the path of publication.

DavidSchinazi:
Fixes #115

### DIFF
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -52,7 +52,7 @@ unacceptable. This document describes a version negotiation mechanism that
 allows a client and server to select a mutually supported version. Optionally,
 if the client's chosen version and the negotiated version share a compatible
 first flight format, the negotiation can take place without incurring an extra
-round trip.
+round trip. This document updates RFC 8999.
 
 
 --- middle
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -10,6 +10,7 @@ wg: QUIC
 number:
 date:
 consensus: true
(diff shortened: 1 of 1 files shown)


LPardue (approved):

tfpauly (approved):

zaheduzzaman:
This PR just updates the tag. Which I think is not adequate. The abstract should clearly mention that it is updating 8999 and the respective sections should say what it is updating.

LPardue:
The whole document updates Section 6 of RFC 8999. I think it's sufficient to state that in a stand-alone sentence in the abstract and introduction.

DavidSchinazi:
As per the IETF tradition, I've modified the abstract to mention that this document updates RFC 8999. There is no requirement to update the introduction or the rest of the text.

nibanks (approved):

zaheduzzaman:
Please merge and update the datatracker version.

PR #131, opened 2022-10-12
Mandate validation of Version Information

### LINKED ISSUE #129: Should fail the handshake when version_info.chosen_version doesn't match long_header.version
Currently at [section 3](https://datatracker.ietf.org/doc/html/draft-ietf-quic-version-negotiation#section-3) says future versions might allow the long header version to be different from the chosen version in the version_info transport parameter.  But there's nothing about what to do until such a version exists which changes the semantics of the chosen_version.
This is in the initial packet, before incompatible version negotiation, and before compatible version negotiation, has happened.

Which error should the handshake be closed with? [Section 4](https://datatracker.ietf.org/doc/html/draft-ietf-quic-version-negotiation#section-4) says that version info transport parameter parsing errors should use the transport parameter error.

DavidSchinazi:
Fixes #129

### DIFF
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -387,7 +387,7 @@ based on information received from a Version Negotiation packet MUST ignore any
 Version Negotiation packets it receives in response to that connection attempt.
 
 Both endpoints MUST parse their peer's Version Information during the handshake.
-If parsing the Version Information failed (for example, if it is too short or if
+If that leads to a parsing failure (for example, if it is too short or if
 its length is not divisible by four), then the endpoint MUST close the
 connection; if the connection was using QUIC version 1, that connection closure
 MUST use a transport error of type TRANSPORT_PARAMETER_ERROR. If an endpoint
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -408,7 +408,7 @@ the Long Headers that are part of the client's first flight is set to
 0x00000001. When the server then parses the client's Version Information, the
 server MUST validate that the client's Chosen Version matches the version in use
 for the connection. If the two differ, the server MUST close the connection with
-a version negotiation error. For example: if a server receives the client's
+a version negotiation error. For example, if a server receives the client's
 Version Information over QUIC version 1 (as indicated by the Version field of
 the Long Header packets that carried the transport parameters) and the client's
 Chosen Version is not set to 0x00000001, the server will close the connection
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -405,7 +405,7 @@ When a server receives a client's first flight, the server will first establish
 which QUIC version is in use for this connection in order to properly parse the
 first flight. For example, the server determines that QUIC version 1 is in use
 by observing that the Version field of the first Long Header packet it receives
-is set to 0x00000001. When the server then parses the client's Version
+is set to 0x00000001. When the server then processes the client's Version
 Information, the server MUST validate that the client's Chosen Version matches
 the version in use for the connection. If the two differ, the server MUST close
 the connection with a version negotiation error. For example, if a server
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -403,16 +403,16 @@ VERSION_NEGOTIATION_ERROR; see {{iana-error}}.
 
 When a server receives a client's first flight, the server will first establish
 which QUIC version is in use for this connection in order to properly parse the
-first flight. For example, QUIC version 1 is in use when the Version field of
-the Long Headers that are part of the client's first flight is set to
-0x00000001. When the server then parses the client's Version Information, the
-server MUST validate that the client's Chosen Version matches the version in use
-for the connection. If the two differ, the server MUST close the connection with
-a version negotiation error. For example, if a server receives the client's
-Version Information over QUIC version 1 (as indicated by the Version field of
-the Long Header packets that carried the transport parameters) and the client's
-Chosen Version is not set to 0x00000001, the server will close the connection
-with a version negotiation error.
+first flight. For example, the server determines that QUIC version 1 is in use
+by observing that the Version field of the first Long Header packet it receives
+is set to 0x00000001. When the server then parses the client's Version
+Information, the server MUST validate that the client's Chosen Version matches
+the version in use for the connection. If the two differ, the server MUST close
+the connection with a version negotiation error. For example, if a server
+receives the client's Version Information over QUIC version 1 (as indicated by
+the Version field of the Long Header packets that carried the transport
+parameters) and the client's Chosen Version is not set to 0x00000001, the server
+will close the connection with a version negotiation error.
 
 If a client receives a Version Information where the server's Chosen Version was
 not sent by the client as part of its Available Versions, the client MUST close
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -716,5 +716,6 @@ codepoint in the 0-63 range to replace the provisional codepoint described above
 # Acknowledgments {#acknowledgments}
 {:numbered="false"}
 
-The authors would like to thank Nick Banks, Mike Bishop, Ryan Hamilton, Roberto
-Peon, Anthony Rossi, and Martin Thomson for their input and contributions.
+The authors would like to thank Nick Banks, Mike Bishop, Martin Duke, Ryan
+Hamilton, Roberto Peon, Anthony Rossi, and Martin Thomson for their input and
+contributions.
--- a/draft-ietf-quic-version-negotiation.md
+++ b/draft-ietf-quic-version-negotiation.md
@@ -392,12 +392,31 @@ its length is not divisible by four), then the endpoint MUST close the
 connection; if the connection was using QUIC version 1, that connection closure
 MUST use a transport error of type TRANSPORT_PARAMETER_ERROR. If an endpoint
 receives a Chosen Version equal to zero, or any Available Version equal to zero,
-it MUST treat it as a parsing failure.
+it MUST treat it as a parsing failure. If a server receives a Version
+Information where the Chosen Version is not included in Available Versions, it
+MUST treat it as a parsing failure.
 
 Every QUIC version that supports version negotiation MUST define a method for
 closing the connection with a version negotiation error. For QUIC version 1,
 version negotiation errors are signaled using a transport error of type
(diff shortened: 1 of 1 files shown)


martinduke (approved):
LGTM

anrossi (approved):

draft-ietf-quic-version-negotiation.md
```diff
@@ -392,13 +392,32 @@ its length is not divisible by four), then the endpoint MUST close the
 connection; if the connection was using QUIC version 1, that connection closure
 MUST use a transport error of type TRANSPORT_PARAMETER_ERROR. If an endpoint
 receives a Chosen Version equal to zero, or any Available Version equal to zero,
-it MUST treat it as a parsing failure.
+it MUST treat it as a parsing failure. If a server receives a Version
```
martinthomson, line 395: What is a parsing failure?  TRANSPORT_PARAMETER_ERROR or something else?

draft-ietf-quic-version-negotiation.md
```diff
@@ -392,13 +392,32 @@ its length is not divisible by four), then the endpoint MUST close the
...
 VERSION_NEGOTIATION_ERROR; see {{iana-error}}.
 
+When a server receives a client's first flight, the server will first establish
+which QUIC version is in use for this connection in order to properly parse the
+first flight. For example, QUIC version 1 is in use when the Version field of
+the Long Headers that are part of the client's first flight is set to
+0x00000001. When the server then parses the client's Version Information, the
+server MUST validate that the client's Chosen Version matches the version in use
+for the connection. If the two differ, the server MUST close the connection with
+a version negotiation error. For example: if a server receives the client's
```
martinthomson, line 411:
```suggestion
a version negotiation error. For example, if a server receives the client's
```

draft-ietf-quic-version-negotiation.md
```diff
@@ -392,13 +392,32 @@ its length is not divisible by four), then the endpoint MUST close the
...
 version negotiation errors are signaled using a transport error of type
 VERSION_NEGOTIATION_ERROR; see {{iana-error}}.
 
+When a server receives a client's first flight, the server will first establish
+which QUIC version is in use for this connection in order to properly parse the
+first flight. For example, QUIC version 1 is in use when the Version field of
+the Long Headers that are part of the client's first flight is set to
+0x00000001. When the server then parses the client's Version Information, the
+server MUST validate that the client's Chosen Version matches the version in use
+for the connection. If the two differ, the server MUST close the connection with
```
martinthomson, line 410: Not sure if this is going to be correct in all cases.

The client might switch to using Version 2 if the server decides to upgrade.  Retransmissions of the ClientHello containing the Version Information might then be sent in Initial packets with a version == 2.

Now, obviously, the server can't upgrade until it has read the Version Information, but there is a possibility that a client could retransmit stuff that the server already has.  This might be due to the Version Information being in a packet that the server received, processed, and acknowledged.  But the ACK might get lost and so the client might still end up sending another copy.

Also, I don't believe that this sort of validation is easy to implement.  The principle we apply here is that validation is mandated only where there is fairly uniform agreement that it is trivial to implement.  This is not.  It requires accessing information that is very hard to access in our stack.  In our stack, processing of Version Information occurs in a callback that is invoked by the TLS library.  For various reasons (mostly relating to how callbacks from C work with Rust and Rust's ownership rules), that processing cannot access a lot of connection-level state.  It certainly can't access details of the packet that was received, even if that would make sense.  After all, the data that the cryptographic handshake processes could come from multiple packets, some of which might have been received and long-since discarded.

draft-ietf-quic-version-negotiation.md
```diff
@@ -392,13 +392,32 @@ its length is not divisible by four), then the endpoint MUST close the
...
+a version negotiation error. For example: if a server receives the client's
+Version Information over QUIC version 1 (as indicated by the Version field of
+the Long Header packets that carried the transport parameters) and the client's
+Chosen Version is not set to 0x00000001, the server will close the connection
+with a version negotiation error.
+
+If a client receives a Version Information where the server's Chosen Version was
+not sent by the client as part of its Available Versions, the client MUST close
+the connection with a version negotiation error.
```
martinthomson, line 419: We didn't say this already?  Wow.  This is definitely needed.

martinthomson (changes requested):
See the long comment for why I think that this is not a good change (though there are definitely good bits).

draft-ietf-quic-version-negotiation.md
DavidSchinazi: That's described earlier in this paragraph

draft-ietf-quic-version-negotiation.md
DavidSchinazi: Agreed, this was assumed by all of us but not actually written down.

draft-ietf-quic-version-negotiation.md
DavidSchinazi: The text above along discusses the client's first flight. The retransmission scenario would not be part of the first flight, by definition. The validation is trivial to implement because you have access to the version of the connection - we're not asking to keep track of the version of the packet carrying the CRYPTO frame, only the version of the first packet received. Your stack already has access to this during TP parsing because the TP parsing is version-dependent.

draft-ietf-quic-version-negotiation.md
```diff
@@ -392,13 +392,32 @@ its length is not divisible by four), then the endpoint MUST close the
 connection; if the connection was using QUIC version 1, that connection closure
 MUST use a transport error of type TRANSPORT_PARAMETER_ERROR. If an endpoint
 receives a Chosen Version equal to zero, or any Available Version equal to zero,
-it MUST treat it as a parsing failure.
+it MUST treat it as a parsing failure. If a server receives a Version
+Information where the Chosen Version is not included in Available Versions, it
+MUST treat it as a parsing failure.
```
ekr, line 397: What's the purpose of this requirement? Is it just a correctness check?

draft-ietf-quic-version-negotiation.md
```diff
@@ -392,13 +392,32 @@ its length is not divisible by four), then the endpoint MUST close the
...
 Every QUIC version that supports version negotiation MUST define a method for
 closing the connection with a version negotiation error. For QUIC version 1,
 version negotiation errors are signaled using a transport error of type
 VERSION_NEGOTIATION_ERROR; see {{iana-error}}.
 
+When a server receives a client's first flight, the server will first establish
+which QUIC version is in use for this connection in order to properly parse the
+first flight. For example, the server determines that QUIC version 1 is in use
+by observing that the Version field of the first Long Header packet it receives
+is set to 0x00000001. When the server then parses the client's Version
```
ekr, line 408: I'm concerned about this "for example". Suppose we defined a QUIC v3 which used the same version field. In S 2.2. we said this was version specific and it has to be for any *pair* of versions.

draft-ietf-quic-version-negotiation.md
DavidSchinazi: Yes, it makes sure that the server validates that the client follows the MUST.

draft-ietf-quic-version-negotiation.md
DavidSchinazi: The QUIC version field contains the QUIC version, so I'm not sure I follow how QUIC version 3 would use the version field of 1.

draft-ietf-quic-version-negotiation.md
ekr: I'm sorry, I don't follow this at all.

draft-ietf-quic-version-negotiation.md
martinthomson: No, it is not.  You really need to use the words "parsing failure" in the definition if you mean to invoke it with this text.

draft-ietf-quic-version-negotiation.md
ekr: NM. I misread this as the client determining the server version. I agree this text is fine.

draft-ietf-quic-version-negotiation.md
DavidSchinazi: In `Client-Sent Available Versions` we have `Note that the version in the Chosen Version field MUST be included in this list`. The text we're commenting on here asks the server to validate that the client followed that MUST

draft-ietf-quic-version-negotiation.md
martinthomson: I think that @ekr is implicitly referring to the potential for the version field to become ossified where we are forced to identify v3 in some other way.

draft-ietf-quic-version-negotiation.md
```diff
@@ -392,13 +392,32 @@ its length is not divisible by four), then the endpoint MUST close the
...
+Information, the server MUST validate that the client's Chosen Version matches
```
martinthomson, line 409: Still opposed to this MUST, for aforementioned reasons.

draft-ietf-quic-version-negotiation.md
martinthomson: But there is no guarantee that the version information is in that first packet, so this is not that easy to implement.

draft-ietf-quic-version-negotiation.md
DavidSchinazi: Sure. Fixed by explicitly using the words "parsing failure" above.

draft-ietf-quic-version-negotiation.md
```diff
@@ -387,18 +387,37 @@ based on information received from a Version Negotiation packet MUST ignore any
...
+the version in use for the connection. If the two differ, the server MUST close
```
martinthomson, line 410: I think that the problem I'm seeing here is rooted in the word "parses" and how it connects to the "in use" bit.  I think that you might want to use "processes" in the first instance.  For the second, use the version of the Initial packet that was used to open the connection rather than a potentially over-broad "in use".

Maybe instead:

> If the server processes Version Information from the client and the Chosen Version differs from the version of the packet that initiated the connection attempt, it MUST close the connection with a version negotiation error.

However, this implies something about Retry packets, so I get why you went with "in use" to avoid that.  Maybe instead:


> If the server processes Version Information from the client and the Chosen Version differs from the version from the version of packets the server has received for this connection attempt, it MUST close the connection with a version negotiation error.

That is, rely on an expectation that the client uses a single, consistent version for its connection attempt.

martinthomson (approved):
Sorry about the confusion.  Language choice lead me astray.  I already implemented this specific check.  I've suggested some wording that might avoid such confusion in future.  Maybe.

martinthomson (approved):
WFM.  David pointed out that this already defines "in use" and the change to "processes" is good.

DavidSchinazi:
@ekr confirmed offline that he can live with this PR as-is, so I'm merging it

PR #132, opened 2022-12-16
Final IANA values

DavidSchinazi:
See [the registry](https://www.iana.org/assignments/quic/quic.xhtml)


martinthomson (approved):
This matches the registry.  17!

